PR #1, opened 2022-03-10
Wsdt/turing monsters

InoMurko:


wsdt (approved):
looks good thanks, but most recent changes are not included yet.

boba_community/turing-monsters/contracts/NFTMonsterV2.sol
```diff
@@ -0,0 +1,121 @@
...
+        address receiver,
+        uint256 royaltyAmount
+    ) {
+        return (address(this), (_salePrice * ROYALTY_PERCENTAGE)/100);
+    }
+
+    /**
+    * For DApp UX to determine how many NFTs can be minted.
+    */
+    function allowedMintAmount(address account_) view external returns (uint256) {
```
souradeep-das: bringing the conversation here - do we specifically need this method? 
since `MAX_MINT_IN_PUBLIC` and `amountMintedInPublicSale()` are both public, the DApp can calculate and show it for the user?

boba_community/turing-monsters/contracts/WithRecover.sol
```diff
@@ -0,0 +1,24 @@
...
+
+contract WithRecover is Ownable {
+
+    /// @dev Recovers ERC20 tokens accidentally sent to the contract.
+    function recoverERC20(address tokenAddress, uint256 tokenAmount) external onlyOwner {
+        IERC20(tokenAddress).transfer(owner(), tokenAmount);
+    }
+
+    /// @dev Recovers ETH accidentally sent to the contract.
+    function recoverGas() external onlyOwner {
```
souradeep-das: > does this compete with the withdraw() on NFTMonsters? if so, can we make it overrideable?

> indirectly it does yes, but since both methods can be executed by the same person this shouldn't make a difference.
But yes that's something we should argue about, might also depend on how the project is organized.

true, in our context, because of the `withdraw()` splitting the balance between all the project owners, having two ways might be a bit equivocal

souradeep-das (approved):

boba_community/turing-monsters/contracts/NFTMonsterV2.sol
wsdt: True that! Will remove it, let's save some gas here.

boba_community/turing-monsters/contracts/WithRecover.sol
wsdt: Sure let's just remove it, the clean way would be just saving how much ETH are minting revenue but this would result in more gas fees.

souradeep-das: aha great point about the total-revenue 👍

PR #8, opened 2022-03-10
Wsdt/bridge onchain metadata

wsdt:


CAPtheorem:
already applied via other PR - can be closed

PR #9, opened 2022-03-10
[L2Geth] Fix the problem of gasLimit > gasUsed

boyuan-chen:
In order to make sure that `tx.Gas() < tx.GasUsed()`, the check code is added in the API level [verfiyFee](https://github.com/bobanetwork/boba/blob/95c28f6661fcd9b311eb66272a46a7b931bca8c9/l2geth/rollup/sync_service.go#L931). 

The calculation of the original l1 security fee is based on `types.Transaction`, but it's very hard to keep track of it in multiple places and make sure that they are the same. We used `tx.Data()` to replace the `types.Transaction`, so l1SecurityFee is the same from the API level and final state transition level. It requires us to increase the overhead fee in `GasOracle` to cover the changes of how we calculate the `l1GasUsed`.

This introduces a new hard fork. To make things easier, we can deploy it with the new fee system.

In estimateGas API, the default gas price is added to correctly calculate l1SecurityFee if the payload doesn't have the gas price option.


l2geth/core/state_transition.go
```diff
@@ -130,16 +130,19 @@ func IntrinsicGas(data []byte, contractCreation, isHomestead bool, isEIP2028 boo
...
-	var isGasUpdate = true
+	isFeeUpdate := evm.ChainConfig().IsFeeUpdate(evm.BlockNumber)
+	isGasUpdate := evm.ChainConfig().IsGasUpdate(evm.BlockNumber)
 	if rcfg.UsingOVM {
 		if msg.GasPrice().Cmp(common.Big0) != 0 {
 			// Compute the L1 fee before the state transition
 			// so it only has to be read from state one time.
 			l1Fee, _ = fees.CalculateL1MsgFee(msg, evm.StateDB, nil)
-			l2ExtraGas, _ = fees.CalculateL2GasForL1Msg(msg, evm.StateDB, nil)
+			if isFeeUpdate {
```
boyuan-chen, line 141: The new gas fork for calculating the l1 security fee based on the `tx.Data()`.

l2geth/internal/ethapi/api.go
```diff
@@ -1056,30 +1056,26 @@ func DoEstimateGas(ctx context.Context, b Backend, args CallArgs, blockNrOrHash
...
-
-	value := new(big.Int)
-	if args.Value != nil {
-		value = args.Value.ToInt()
-	}
-
 	gasPrice := new(big.Int)
 	if args.GasPrice != nil {
 		gasPrice = args.GasPrice.ToInt()
+	} else {
```
boyuan-chen, line 1058: The default gas price is added to return the correct l1 security fee if the gas price is not provided in the payload.

l2geth/rollup/sync_service.go
```diff
@@ -1001,6 +1006,29 @@ func (s *SyncService) verifyFee(tx *types.Transaction) error {
...
+			return err
+		}
+		// Ensure that tx.Gas() is larger than l1SecurityFee / l2GasPrice + intrGas
+		l2ExtraGas, err := fees.CalculateL1GasFromGPO(tx.Data(), l2GasPrice, gpo)
+		if err != nil {
+			return err
+		}
+		estimateGas := new(big.Int).Add(l2ExtraGas, big.NewInt(int64(intrGas)))
+		log.Debug("Validate gas limit", "estimateGas", estimateGas, "l2GasPrice", l2GasPrice, "gasLimit", tx.Gas(), "intrGas", intrGas, "l2ExtraGas", l2ExtraGas)
+		if big.NewInt(int64(tx.Gas())).Cmp(estimateGas) < 0 {
```
boyuan-chen, line 1025: Make sure that `tx.Gas()` > `tx.GasUsed()`

mmontour1306 (approved):

CAPtheorem (approved):

Issue #13, opened 2022-03-14
ERC721 Metadata and NFT Bridges


souradeep-das:
Categorizing ERC721s into three types on the basis of metadata, as per our interests

1.  ERC721 with derivable metaData (more common)
`tokenURI` returns in the form `baseURI` + `tokenID`
2.  ERC721 with custom metaData
not 1) and`tokenURI` is unique for each token (non-derivable on-chain) or is dependent on custom salt or fully on-chain metadata
3. ERC721 with no metadata (non- ERC721Metadata)

Currently, we can support 1). 
With the StandardERC721 representation, the `baseURI` can be specified while deploying the representation on the opposite layer. so, metadata can be derivable for representation ERC721s on the other side

We also support 3). However - the representation ERC721 would still be ERC721Metadata compliant and would return "" for `tokenURI`, which we can avoid by having a second plain ERC721Standard(without metaData interfaces). Associated with #12, checks to make sure both original-representation contracts comply to same ERC165 ids can be made on registerPair()

We do not have support for 2). yet. Some forms of these - which do not confirm to the type 1) but still can be fully derived with existing data - can be supported with the customERC721 here - #8

Other forms - where the metadata generated depends on unique user/oracle inputs (for example Turing Monsters) might require - sending this extra data between layers ( just so because this metadata cannot be generated one the representation ERC721) However, To make the bridge support generally all-kinds of these ERC721s we need to send the tokenURI between the layers (since the bridge cannot or doesn't know how to access the unique salts)

Approach -1 
could be- If the ERC721 confirms to ERC721Metadata, get the `tokenURI` and add a new param while sending the xDomain message. If it doesn't send "". On the message receiving end -make the bridge: mint + setTokenURI on the representation ERC721.

Approach -2
To further reduce gas, since most of the ERC721s confirm to the baseURI+tokenId, format - sending over tokenURI for each could just be wasting gas. With this approach we separate and provide choice to either bridge With/Without Metadata. and we keep baseURI on representation contract to allow resolving

|       | baseURI contracts           | non baseURI (onchain)  |
| ------------- |:-------------:| -----:|
| **with Metadata**    | overriden by baseURI | sets to actual metadata |
| **without Metadata**     | baseURI handles data      |   sets to empty metadata |

The problem though is with sending over the tokenURI of contracts like (TuringMonsters - which have onchain metadata and rely on inputs/salt etc to recreate the metadata)

[...]

PR #21, opened 2022-03-17
NFT: add bridge_with_extraData option

souradeep-das:
if your NFTs need to communicate extra data to the other layer - expose all the extra data encoded together in a method bridgeExtraData , the bridge will be able to prioritise and pick that data up while bridging

associated with #13

approach followed-
https://github.com/bobanetwork/boba/issues/13#issuecomment-1068070777

1. For these specialized ERC721s in discussion have the requirement of a special method called 'bridgeExtraData()' to enable metadata or data transfer between layers. If there is a need for multiple seed/data to be transferred, the bridgeExtraData() needs to expose the final encoded data that will be sent along.
2. Provide this option to the users - to, for instance, withdrawWithExtraData() on the NFTBridges - where if the contract declares an bridgeExtraData() method, use the encoded data on the xDomain message, if not try sending the tokenURI. This works as an override
3. the handling of this extraData could depend on the representation contract on the other side, extraData can be passed along with the mint() method

bridging erc721s with common metadata format "baseURI" + tokenId (or) no metadata - does not need to worry about this.
This could allow support for cases when the metadata is not in the form above (or) depends on external unique inputs for generation

for ref: associated changes 
https://github.com/bobanetwork/boba/commit/c56dd027cb8ac81ae2278408290c51d5627b78f3 #20 

(tests to be added)


packages/boba/contracts/contracts/bridges/L1NFTBridge.sol
```diff
@@ -207,16 +208,76 @@ contract L1NFTBridge is iL1NFTBridge, CrossDomainEnabled, ERC721Holder, Reentran
...
+        bytes memory extraData;
+        // if token has base on this layer
+        if (pairNFTInfo[_l1Contract].baseNetwork == Network.L1) {
+            // check the existence of bridgeExtraData(uint256) on l1Contract
+            if (ERC165Checker.supportsInterface(_l1Contract, 0x9b9284f9)) {
+                extraData = iSupportBridgeExtraData(_l1Contract).bridgeExtraData(_tokenId);
+            } else {
+                // otherwise send tokenURI return (encoded in bytes)
+                // allow to fail if the call fails
+                extraData = abi.encode(IERC721Metadata(_l1Contract).tokenURI(_tokenId));
```
wsdt, line 251: @souradeep-das So this means we now simply send the whole tokenURI to L1 and save it there probably in a mapping or so?

souradeep-das: the first priority is the return from the bridgeExtraData() method, so if turingMonster or a similar contract will implement this optional method, the bridge will send the `generation` data instead, reference: https://github.com/bobanetwork/boba/commit/c56dd027cb8ac81ae2278408290c51d5627b78f3

if a similar contract has 2,3 generation data - the method can expose a encoded bytes similarly

if a contract doesn't implement the special method, `depositNFTWithExtraData()`defaults to sending data from tokenURI - (this could be used by someone whose erc721s have distinct tokenURIs)

wsdt:
@souradeep-das saw the integration test contracts. The actual integration tests haven't been added so far?

packages/boba/contracts/contracts/standards/L1StandardERC721.sol
```diff
@@ -43,7 +43,7 @@ contract L1StandardERC721 is IL1StandardERC721, ERC721 {
         return _interfaceId == firstSupportedInterface || _interfaceId == secondSupportedInterface;
     }
 
-    function mint(address _to, uint256 _tokenId) public virtual override onlyL1Bridge {
+    function mint(address _to, uint256 _tokenId, bytes memory _data) public virtual override onlyL1Bridge {
```
souradeep-das, line 44: should we emit the `_data` here with events on `mint()`?
(to cover unused)

packages/boba/contracts/contracts/bridges/L2NFTBridge.sol
```diff
@@ -231,22 +232,96 @@ contract L2NFTBridge is iL2NFTBridge, CrossDomainEnabled, ERC721Holder, Reentran
...
+            // check the existence of bridgeExtraData(uint256) on l2Contract
+            if (ERC165Checker.supportsInterface(_l2Contract, 0x9b9284f9)) {
+                extraData = iSupportBridgeExtraData(_l2Contract).bridgeExtraData(_tokenId);
+            } else {
+                // otherwise send tokenURI return (encoded in bytes)
+                // allow to fail if the call fails
+                extraData = abi.encode(IERC721Metadata(_l2Contract).tokenURI(_tokenId));
+            }
+        }
+        // size limits unchecked
```
souradeep-das, line 285: should we check the size of the `extraData` here? to limit the size of `data` that can be communicated 
currently, the erc721s registered with the bridge are usable, so shouldn't be a problem as long as registrations are controlled?

InoMurko:
A general question - `withdrawWithExtraData`. Why wouldn't you withdraw with "extra" data?

It seems to me that if you're bridging a NFT from a layer to a layer, you're always moving the whole NFT. Every single bit of information is a NFT, if you're not moving everything, it's something else, it's not *that* NFT anymore.

Why do we leave the option of perhaps not moving the whole NFT with it's metadata? What case is there where you'd "leave" metadata behind?

CAPtheorem:
Yes leaving metadata behind would seem to be a dead end? So rename `withdrawWithExtraData`->`withdraw` and remove the old `withdraw`?

souradeep-das:
almost all of the general NFTs confirm to having the metadata as a form of - "baseURI" + "tokenId"
so it will be a waste of gas to migrate the metadata for each of such tokens (which confirm to the general standard)

the tokenURI for such token can simply be generated by having the baseURI set once on the contract - enabled by commit https://github.com/omgnetwork/optimism-v2/commit/bced62f12739ef968db3ad7c00707fca26b61900

so, generally for all normal NFTs - users can just continue using the withdraw() method, which is cheaper - and will still be a complete representation
getting to this approach is outlined here: `approach-2` from https://github.com/bobanetwork/boba/issues/13#issue-1168406643  
lmk what you think about this

CAPtheorem (approved):

PR #22, opened 2022-03-17
Support Boba token as the fee token

boyuan-chen:
## Changes:
- Rewrite the logic of calculating the gas limit for l2geth
  Rewriting this whole logic makes sure that the l1SecurityFee is correctly calculated in the gas usage and `tx.GasLimit()` is larger or equal to the `tx.GasUsed()`. Also, a precaution for the situation of the l1SecurityFee is too large to cause that the gas limit is larger than the block gas limit is added.
  > In order to add `l1SecurityFee` to the gas usage, the code in `api.go` and `state_transition.go` are modified. When users call `eth_estimateGas`, the `DoEstimateGas` function in `api.go` calls `TransitionDb` function in `state_transition.go` by using the binary search to find an executable gas limit. In `state_transition.go`, we calculate the gas usage for the l1SecurityFee if the `gas price` is provided. However, there are situations that `gas price` is not provided, which means that the returned gas limit doesn't consider the l1SecurityFee. To solve this problem, the code of overriding the gas price is added in `DoEstimateGas` function to secure that the gas limit covers the `l1SecurityFee`. This introduces a new problem for the local testing environment. We allow **0** gas price in the local testing environment. If you are trying to send all ETH from your wallet to another wallet using the zero gas price without the gas limit, the `DoEstimateGas` function will fail. This is caused by overriding the gas price, so the total ETH cost is larger than the actual ETH amount in your wallet. In the production environment, only gas oracle owner can use the zero gas price and we allow this configuration in `DoEstimateGas`, so this is not a problem for the production environment.

  > There are two different cases that the gas limit might be too low and returns errors. The first one is that the gas limit is below `IntrinsicGas + l1SecurityFee / l2GasPrice`. In this case, the tx can't pass the `verifyFee` function in `sync_service.go` and returns `intrinsic gas too low` and we don't charge the tx fee for this. Another one is to override the estimated gas value from `eth_estimateGas` by giving a smaller value. Since users are confident with this value, we revert the tx and charge the tx fee. Also, this protects us from the DOS attack.
  
- Add Boba as the fee token in l2geth
  The gas usages for the **Boba fee token** and **ETH fee token** are the same. The transaction fee for the **Boba fee token** is to multiple `priceRatio = ETHPrice / BobaPrice` and the gas usage.
- Add `Boba_GasPriceOracle.sol` contract to collect fees and configure settings for using the boba as the fee token
  Users can select the fee token in this contract by calling `useBobaAsFeeToken` and `UseETHAsFeeToken`. The default option is to use ETH as the fee token

## IMPORTANT

The l2 gas price and l1 base price should be calculated carefully. If we go with the `l2GasPrice=1GWEI`, the minimum Boba amount for using the Boba as the fee token is around 11 Boba. This is caused by how the gas limit is estimated. In `api.go`, it estimates the gas from the middle of the block gas limit and uses the binary search to find a good gas limit. The minimum Boba for the `block.GasLimit / 2` is `gasLimit * gasPrice * priceRatio = 11_000_000 / 2 * 10^9 * 2000 / 10^18 = 11 BOBA`. The ideal l2 gas price should be 0.1 GWEI, so the minimum Boba for users to use the Boba as the fee token is 1.1 BOBA

## TODO:

- [x] Add boba fee cost in `tx.Receipt()` and tests for it
- [x] Should calculate the gas usage for different gas price
- [x] Add gas usage tests in l2geth
- [x] Add precaution tests for the situation that the l1SecurityFee is too large to break the gas limit
- [x] Add tests for configuring the fee tokens in `Boba_GasPriceOracle.sol`


l2geth/internal/ethapi/api.go
```diff
@@ -1005,20 +1007,47 @@ func DoEstimateGas(ctx context.Context, b Backend, args CallArgs, blockNrOrHash
...
+	state, _, _ := b.StateAndHeaderByNumberOrHash(ctx, blockNrOrHash)
+	// Use non-zero gas price for the l1 security fee if gas price is not nil
+	// In production, users have to pay the gas fee. In the local testing system,
+	// gas price can be zero. If their balance is really low for using zero gas price,
+	// they can't bypass this estimateGas
+	gasPrice := new(big.Int)
+	price, err := b.SuggestPrice(ctx)
+	if err == nil && isFeeTokenUpdate {
+		// Override gas price
+		if args.GasPrice == nil {
```
boyuan-chen, line 1024: Override the `nil` gas price, so we can estimate gas limit with the l1 security fee in `state_transition.go`

l2geth/core/state_processor.go
```diff
@@ -117,6 +132,14 @@ func ApplyTransaction(config *params.ChainConfig, bc ChainContext, author *commo
 		return nil, err
 	}
 
+	// Get the l2 boba fee if users use Boba token as the fee token
+	L2BobaFee := new(big.Int)
```
boyuan-chen, line 140: Insert l2 boba fee into the receipt for the block explorer

l2geth/rollup/sync_service.go
```diff
@@ -949,6 +945,18 @@ func (s *SyncService) verifyFee(tx *types.Transaction) error {
...
+
+	// Check if the wallet address picks BOBA as the fee token
+	feeTokenSelection := state.GetFeeTokenSelection(from)
+	isBobaFeeTokenSelect := feeTokenSelection.Cmp(common.Big1) == 0 && isFeeTokenUpdate
+
+	// Prevent transactions without enough balance from
+	// being accepted by the chain but allow through 0
+	// gas price transactions
+	cost := tx.Value()
+	if !isBobaFeeTokenSelect && tx.GasPrice().Cmp(common.Big0) != 0 {
```
boyuan-chen, line 957: Make sure that we don't add extra fee if users pick Boba as the fee token

packages/contracts/contracts/L2/predeploys/Boba_GasPriceOracle.sol
```diff
@@ -0,0 +1,165 @@
...
+     * Get Boba cost for cost estimation
+     * @param _txData the data payload
+     */
+    function getBobaCost(bytes memory _txData) public view returns (uint256) {
+        OVM_GasPriceOracle gasPriceOracleContract = OVM_GasPriceOracle(gasPriceOracleAddress);
+        return gasPriceOracleContract.getL1Fee(_txData) * priceRatio;
+    }
+
+    /**
+     * withdraw BOBA tokens to l1 fee wallet
```
CAPtheorem, line 231: onlyOwner? (Although I guess its not a security risk since the targets are hardcoded?)

packages/contracts/contracts/L2/predeploys/Boba_GasPriceOracle.sol
```diff
@@ -0,0 +1,165 @@
...
+     *************/
+
+    // Address on L1 that will hold the fees once withdrawn. Dynamically initialized within l2geth.
+    address public l1FeeWallet;
+
+    // L2 Boba token address
+    address public l2BobaAddress;
+
+    // The maximum value of ETH and BOBA
+    uint256 public maxPriceRatio = 5000;
```
CAPtheorem, line 43: this is all fine since can be changed if needed

boba_examples/boba-fee/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@boba/boba-fee",
+  "version": "1.0.0",
+  "scripts": {
+    "use:boba": "FEE_TOKEN=BOBA node ./src/index.js",
+    "use:eth": "FEE_TOKEN=ETH node ./src/index.js"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "@eth-optimism/contracts": "^0.5.18",
```
CAPtheorem: This will trigger using contracts from outside the boba monorepo? The contracts version inside the monorepo is 0.5.11?

l2geth/core/state_transition.go
```diff
@@ -330,6 +395,13 @@ func (st *StateTransition) refundGas() {
 	remaining := new(big.Int).Mul(new(big.Int).SetUint64(st.gas), st.gasPrice)
```
CAPtheorem: These next two lines are basically always zero now?

l2geth/core/state_transition.go
```diff
@@ -330,6 +395,13 @@ func (st *StateTransition) refundGas() {
...
 	st.state.AddBalance(st.msg.From(), remaining)
```
CAPtheorem: would write this as if() else so its clear that `st.state.AddBalance(st.msg.From(), remaining)` handles the normal ETH case. But is that always zero?

CAPtheorem (approved):
Assume this is all automatically compatible with the replicas and the verifier since the fee-decision is set in the contract rather than being set by the user on a per TX basis - good design!

l2geth/core/state_transition.go
boyuan-chen: No, it's zero only when users use Boba as the fee token.
```
if isBobaFeeTokenSelect {
  gasPrice = big.NewInt(0)
  bobaPriceRatio = evm.StateDB.GetBobaPriceRatio()
}
```

packages/contracts/contracts/L2/predeploys/Boba_GasPriceOracle.sol
boyuan-chen: The target address is hardcoded like our ETH vault contract.

boba_examples/boba-fee/package.json
boyuan-chen: Removed the `@eth-optimism/contracts` package

integration-tests/test/boba-fee-payment.spec.ts
```diff
@@ -0,0 +1,793 @@
...
+    )
+
+    Proxy__Boba_GasPriceOracle =
+      await Factory__Proxy__Boba_GasPriceOracle.deploy(
+        Boba_GasPriceOracle.address
+      )
+    await Proxy__Boba_GasPriceOracle.deployTransaction.wait()
+  })
+
+  it('{tag:other} should register to use boba as the fee token', async () => {
```
InoMurko: maybe `boba` is a better tag, as it's a `boba` specific feature

packages/contracts/bin/take-dump.ts
```diff
@@ -108,6 +111,7 @@ import { makeL2GenesisFile } from '../src/make-genesis'
     l1CrossDomainMessengerAddress,
     bobaTuringPrice,
     TuringHelperJson,
+    l1BobaTokenAddress,
```
InoMurko: we should add this after `berlinBlock` so that it makes more timeline sense.

packages/boba/contracts/deploy/019-Dump.deploy.ts
```diff
@@ -12,6 +12,10 @@ const deployFn: DeployFunction = async (hre) => {
 
   const deployments = await hre.deployments.all()
 
+  const addressManager = getContractFactory('Lib_AddressManager')
```
InoMurko: why is this added? it's not used

InoMurko:
If I understand this right, the way Boba can be used as a fee token is that you first register your decision on a predeploy contract?

integration-tests/test/boba-fee-payment.spec.ts
```diff
@@ -0,0 +1,793 @@
...
+
+  it('{tag:other} should not register the fee tokens for non EOA accounts', async () => {
+    await expect(Proxy__Boba_GasPriceOracle.useBobaAsFeeToken()).to.be.reverted
+    await expect(Proxy__Boba_GasPriceOracle.useETHAsFeeToken()).to.be.reverted
+  })
+
+  it('{tag:other} Paying a nonzero but acceptable boba gasPrice fee for transferring ETH', async () => {
+    await setPrices(env, 1000)
+
+    const amount = utils.parseEther('0.0000001')
```
InoMurko, line 84: are these two tests coupled? 
```
it('{tag:other} should register to use boba as the fee token', async () => {
it('{tag:other} Paying a nonzero but acceptable boba gasPrice fee for transferring ETH', async () => {
```

boyuan-chen: Yes, we make sure that the Boba fee is correctly charged when users transfer ETH, BOBA and deploy contracts.

mmontour1306:
Two of the Turing tests are broken (local network). Possibly related to zero-gas-price logic which is used by the Turing cache to determine whether or not a call is allowed to block the calling thread.

  1) Stableswap at AWS Lambda
       should correctly swap X in for Y out:
     Error: cannot estimate gas; transaction may fail or may require manual gas limit (error={"reason":"processing response error","code":"SERVER_ERROR","body":"{\"jsonrpc\":\"2.0\",\"id\":132,\"error\":{\"code\":-32000,\"message\":\"execution reverted: TURING: Missing cache entry\"}}\n","error":{"code":-32000},"requestBody":"{\"method\":\"eth_estimateGas\",\"params\":[{\"gas\":\"0x2dc6c0\",\"from\":\"0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266\",\"to\":\"0xc9a43158891282a2b1475592d5719c001986aaec\",\"data\":\"0x94e6c7cb0000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000003c68747470733a2f2f6939697a6e6d6f3333652e657865637574652d6170692e75732d656173742d312e616d617a6f6e6177732e636f6d2f737761707900000000\"}],\"id\":132,\"jsonrpc\":\"2.0\"}","requestMethod":"POST","url":"http://localhost:8545"}, method="estimateGas", transaction={"from":"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266","gasLimit":{"type":"BigNumber","hex":"0x2dc6c0"},"to":"0xC9a43158891282A2B1475592D5719c001986Aaec","data":"0x94e6c7cb0000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000003c68747470733a2f2f6939697a6e6d6f3333652e657865637574652d6170692e75732d656173742d312e616d617a6f6e6177732e636f6d2f737761707900000000","accessList":null}, code=UNPREDICTABLE_GAS_LIMIT, version=providers/5.5.3)

  2) Pull Bitcoin - USD quote
       should get the current Bitcoin - USD price:
     Error: cannot estimate gas; transaction may fail or may require manual gas limit (error={"reason":"processing response error","code":"SERVER_ERROR","body":"{\"jsonrpc\":\"2.0\",\"id\":117,\"error\":{\"code\":-32000,\"message\":\"execution reverted: TURING: Missing cache entry\"}}\n","error":{"code":-32000},"requestBody":"{\"method\":\"eth_estimateGas\",\"params\":[{\"gas\":\"0x2dc6c0\",\"from\":\"0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266\",\"to\":\"0x4b6ab5f819a515382b0deb6935d793817bb4af28\",\"data\":\"0x5d57384a000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000003c68747470733a2f2f6939697a6e6d6f3333652e657865637574652d6170692e75732d656173742d312e616d617a6f6e6177732e636f6d2f71756f74650000000000000000000000000000000000000000000000000000000000000000000000074254432f55534400000000000000000000000000000000000000000000000000\"}],\"id\":117,\"jsonrpc\":\"2.0\"}","requestMethod":"POST","url":"http://localhost:8545"}, method="estimateGas", transaction={"from":"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266","gasLimit":{"type":"BigNumber","hex":"0x2dc6c0"},"to":"0x4b6aB5F819A515382B0dEB6935D793817bB4af28","data":"0x5d57384a000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000003c68747470733a2f2f6939697a6e6d6f3333652e657865637574652d6170692e75732d656173742d312e616d617a6f6e6177732e636f6d2f71756f74650000000000000000000000000000000000000000000000000000000000000000000000074254432f55534400000000000000000000000000000000000000000000000000","accessList":null}, code=UNPREDICTABLE_GAS_LIMIT, version=providers/5.5.3)

packages/contracts/contracts/L2/predeploys/Boba_GasPriceOracle.sol
```diff
@@ -0,0 +1,165 @@
...
+        require(_gasPriceOracleAddress != address(0));
+        gasPriceOracleAddress = _gasPriceOracleAddress;
+        emit UpdateGasPriceOracleAddress(owner(), _gasPriceOracleAddress);
+    }
+
+    /**
+     * Get Boba cost for cost estimation
+     * @param _txData the data payload
+     */
+    function getBobaCost(bytes memory _txData) public view returns (uint256) {
```
souradeep-das: this should only get the 'l1' part of the fee? should we rename it to avoid confusion?

souradeep-das (commented):
looks nice to me!

packages/contracts/contracts/L2/predeploys/Boba_GasPriceOracle.sol
boyuan-chen: Yes!

PR #26, opened 2022-03-21
Wsdt/turing monsters

wsdt:


boba_community/turing-monsters/contracts/NFTMonsterV2Flat.sol
```diff
@@ -0,0 +1,1841 @@
+// SPDX-License-Identifier: MIT
```
souradeep-das, line 1: the flattened contract seems to be outdated, could you run flatten once before merge?

boba_community/turing-monsters/contracts/AddLimitedSupply.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: MIT
```
souradeep-das, line 1: can remove the With'LimitedSupply/Recover/onChain..' contracts if they are unused  :)

souradeep-das (approved):
🚀

CAPtheorem (approved):

PR #32, opened 2022-03-22
Gateway: enable WAMGI; Monster Mint; new SDK

CAPtheorem:


packages/boba/gateway/src/containers/wallet/token/tokenList/TokenList.js
```diff
@@ -247,7 +248,7 @@ function TokenList({
                     aria-label="WAGMIv1"
                   />
                   <Button
-                    onClick={() => { settle_v1() }}
+                    onClick={() => { doSettle_v1() }}
                     variant="contained"
                     disabled={true}
```
sk-enya, line 253: @CAPtheorem  ByDefault **_settle_** button is disabled but as it seems like it's only supported on **_L2 & mainnet_** so can't we use the same condition to enable and disable the settle button?

sk-enya: as for other below.

sk-enya (approved):

packages/boba/gateway/src/containers/wallet/token/tokenList/TokenList.js
CAPtheorem: The settle will be turn on on specific dates in the future. So just hardcoded `disable` until those future dates

PR #41, opened 2022-03-24
BOBA fee minor smoothing; more comments, minor tweeks

CAPtheorem:
Also tried to simplify the logic to make it more clear what is being charged when

### DIFF
--- a/l2geth/core/state_transition.go
+++ b/l2geth/core/state_transition.go
@@ -172,9 +172,9 @@ func NewStateTransition(evm *vm.EVM, msg Message, gp *GasPool) *StateTransition 
 	}
 
 	return &StateTransition{
-		gp:          gp,
-		evm:         evm,
-		msg:         msg,
+		gp:  gp,
+		evm: evm,
+		msg: msg,
 		// gasPrice is normally set via msg.GasPrice() but we override
 		// gasPrice to pass information about the fee choice
 		gasPrice:    gasPrice,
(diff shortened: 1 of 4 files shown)


l2geth/core/state_transition.go
```diff
@@ -147,7 +147,8 @@ func NewStateTransition(evm *vm.EVM, msg Message, gp *GasPool) *StateTransition
 	// The default fee token is ETH
 	isBobaFeeTokenSelect := false
 	if rcfg.UsingOVM {
-		if msg.GasPrice().Cmp(common.Big0) != 0 {
+		// if msg.GasPrice > 0, then...
+		if msg.GasPrice().Cmp(common.Big0) == 1 {
```
boyuan-chen, line 151: We only add l1 security fee when the gas price is not equal to 0

CAPtheorem: Yes, msg.GasPrice().Cmp(common.Big0) == 1 is when msg.GasPrice > 0.

CAPtheorem: Do we ever use a negative gasPrice?

boyuan-chen: No, the negative gas price can't be hexlified. The gasPrice is not a fixed value, we can adjust it.

boyuan-chen: `gasPrice` is the l2GasPrice in `OVM_GasPriceOracle.sol` contract, so it can be any value.

CAPtheorem: ok, so in that case, either `msg.GasPrice().Cmp(common.Big0) == 1` or `msg.GasPrice().Cmp(common.Big0) != 0` is the same thing since .Cmp returns 1 when GasPrice() > 0. Either way is good

boyuan-chen: Yes!

boyuan-chen (approved):

PR #49, opened 2022-03-28
Add meta transaction for selecting Boba as the fee token

boyuan-chen:
Users can bridge only Boba token and use the meta transaction to select the Boba as the fee token


integration-tests/test/boba-fee-payment.spec.ts
```diff
@@ -790,4 +792,239 @@ describe('Boba Fee Payment Integration Tests', async () => {
...
+      const data: any = {
+        primaryType: 'Permit',
+        types: { EIP712Domain, Permit },
+        domain: { name, version, chainId, verifyingContract },
+        message: { owner, spender, value, nonce, deadline },
+      }
+
+      const signature = ethSigUtil.signTypedData(
+        Buffer.from(env.l2Wallet_2.privateKey.slice(2), 'hex'),
+        { data }
```
souradeep-das, line 840: should we use `TypedDataVersion.V4`?

integration-tests/test/boba-fee-payment.spec.ts
```diff
@@ -790,4 +792,239 @@ describe('Boba Fee Payment Integration Tests', async () => {
...
+          value,
+          deadline,
+          sig.v,
+          sig.r,
+          sig.s
+        )
+      ).to.be.revertedWith('execution reverted: ERC20Permit: invalid signature')
+    })
+
+    it("{tag:boba} should revert transaction if users don't have insufficient Boba token", async () => {
```
souradeep-das: minor: typo - change to 'sufficient'

souradeep-das (approved):

packages/contracts/contracts/L2/predeploys/Boba_GasPriceOracle.sol
```diff
@@ -88,6 +96,30 @@ contract Boba_GasPriceOracle is Ownable {
         emit UseBobaAsFeeToken(msg.sender);
     }
 
+    /**
+     * Add the users that want to use BOBA as the fee token
+     * using the Meta Transaction
+     * NOTE: Only work for the mainnet and local testnet
```
InoMurko:
```suggestion
     * NOTE: Only works for the mainnet and local testnet
```

InoMurko (approved):

integration-tests/test/boba-fee-payment.spec.ts
boyuan-chen: There are two packages. `@metamask/eth-sig-util` supports to select the version and communicate with MetaMask, but somehow it doesn't work in the our integration test script. `eth-sig-util` works perfectly. In gateway, we should use `@metamask/eth-sig-util`.
```
  const signature = ethSigUtil.signTypedData(
    {privateKey: Buffer.from(L2Wallet.privateKey.slice(2), 'hex'),
    data,
    version: ethSigUtil.SignTypedDataVersion['V4'],
  })
```

PR #51, opened 2022-03-29
Support proxy contract for Boba_GasPriceOracle

boyuan-chen:
Support proxy contract for Boba_GasPriceOracle


CAPtheorem (approved):

packages/contracts/contracts/L2/predeploys/Boba_GasPriceOracle.sol
```diff
@@ -81,12 +98,51 @@ contract Boba_GasPriceOracle is Ownable {
         require(_l1FeeWallet != address(0) && _l2BobaAddress != address(0));
```
souradeep-das, line 98: let's remove the constructor and the `onlyOwner` if we want to have it under a proxy?

PR #76, opened 2022-04-07
fix for poll interval + theme in ecosystem.

sk-enya:
fix for poll interval + theme


CAPtheorem:
will handle via dedicated PR

PR #83, opened 2022-04-08
Add swapBOBAForETHMetaTransaction

boyuan-chen:
- [x] Update `Boba_GasPriceOracle` contract
- [x] Add unit tests
- [x] Add integration tests
- [x] Update API endpoint
- [x] Update graph node


CAPtheorem (approved):

packages/contracts/contracts/L2/predeploys/Boba_GasPriceOracle.sol
```diff
@@ -139,7 +163,7 @@ contract Boba_GasPriceOracle {
      * using the Meta Transaction
      * NOTE: Only works for the mainnet and local testnet
      */
-    function useBobaAsFeeTokenMetaTransaction(
+    function swapBOBAForETHMetaTransaction(
```
souradeep-das, line 171: i personally think the previous name for this method was better. is the new name kinda confusing?

packages/contracts/test/contracts/L2/predeploys/Boba_GasPriceOracle.spec.ts
```diff
@@ -398,6 +398,100 @@ describe('Boba_GasPriceOracle', () => {
...
+          receivedETHAmount
+        )
+      )
+        .to.emit(Boba_GasPriceOracle, 'UpdateReceivedETHAmount')
+        .withArgs(await signer1.getAddress(), receivedETHAmount)
+    })
+  })
+
+  describe('get receivedETHAmount', () => {
+    it('should revert if caller is not owne', async () => {
```
souradeep-das:
```suggestion
    it('should revert if caller is not owner', async () => {
```

souradeep-das (approved):
🚀

packages/contracts/contracts/L2/predeploys/Boba_GasPriceOracle.sol
boyuan-chen: We provide a small swap for users that don't have enough ETH when they switch fee token to Boba.

souradeep-das: but is this is also the useBobaAsFeeToken meta transaction?

boyuan-chen: No, we don't need another meta transaction for switching the fee token. Even users switch to Boba as the fee token, they have to have some ETH in their wallets. MetaMask checks the user's balance and blocks the tx if there is no ETH in their wallet.

The logic for normal users that only bridge BOBA to L2 is that they use this meta transaction to swap some ETH and then change fee token to BOBA using the normal tx. They have some ETH in their wallet and also they use the BOBA as the fee token.

souradeep-das: got it, alright!

PR #97, opened 2022-04-12
Additional Turing Example: Faucet/CAPTCHA

CAPtheorem:
Please see if we can generate better error messages when people don't solve the captcha correctly


wsdt:
Added a better error message

wsdt:
ToDo:

```
so main things are 
(1) clarify that people need Rinkeby ETH in the first place otherwise they can't do anything 
(2) when user rejects transaction, show the rror message and stop the spinner, and 
(3) deal with the reject -transaction bug , and 
(4) intercept the error message from the smart contract and replace it with "CAPTCHA incorrect" in the front end

basically, if they reject the metamask transaction it generates an error because message is zero or something like that

I think it keeps waiting for the server resonse
```

**So things to do: (1) warning message if no ETH, (2) stop loading if MetaMask tx reject or if TX fails, and (3) redeploy the changed parts**

PR #101, opened 2022-04-13
fix: sdk - remove dependence on l2 eth_getLogs

souradeep-das:
for the relayer, get messageStatus from the L1 contracts directly instead of tracking logs


packages/message-relayer/src/service.ts
```diff
@@ -239,7 +239,8 @@ export class MessageRelayerService extends BaseService<MessageRelayerOptions> {
             // Filter out messages which have been processed
             const newMB = []
             for (const cur of this.state.messageBuffer) {
-              const status = await this.state.messenger.getMessageStatus(cur)
+              const status =
+                await this.state.messenger.getMessageStatusFromContracts(cur)
```
souradeep-das: since we only need succeded/failed status here, the finalization(state batch submission) log binary search can be avoided if needed. lmk if we want to avoid going through the complete search once more here

packages/message-relayer/src/service.ts
```diff
@@ -442,9 +444,10 @@ export class MessageRelayerService extends BaseService<MessageRelayerOptions> {
                 }
               }
 
-              const status = await this.state.messenger.getMessageStatus(
-                message
-              )
+              const status =
+                await this.state.messenger.getMessageStatusFromContracts(
+                  message
+                )
```
souradeep-das, line 469: since we only need succeded/failed status here, the finalization(state batch submission) log binary search can be avoided if needed. lmk if we want to avoid going through the complete search once more here

packages/message-relayer/src/service.ts
CAPtheorem: Please pick whatever makes most sense

packages/message-relayer/src/service.ts
CAPtheorem: Please pick whatever makes most sense

InoMurko (approved):

PR #117, opened 2022-04-17
Gateway Visuals

CAPtheorem:


packages/boba/gateway/src/layout/index.js
```diff
@@ -127,27 +138,31 @@ function App () {
             props: { variant: 'contained', color: 'primary' },
             style: {
               // background: 'linear-gradient(131.81deg, #4A6FEF 2.66%, #4251F0 124.21%)',
-              background: radioGreen,
+              background: buttonColor,
```
sk-enya, line 136: @CAPtheorem  seems like button colour changed to blue which is not looking good. is this intentional ?

PR #122, opened 2022-04-19
Separate prod and dev env for batch submitter

boyuan-chen:
Separate production and development environments


ops/scripts/batch-submitter.sh
```diff
@@ -37,6 +38,7 @@ curl \
     --retry-delay 3 \
     --output /dev/null \
     $BATCH_SUBMITTER_KMS_ENDPOINT
```
InoMurko, line 39: This chunk-check could actually be removed!

boyuan-chen: Removed.

go/batch-submitter/batch_submitter.go
```diff
@@ -117,6 +117,13 @@ func Main(gitVersion string) func(ctx *cli.Context) error {
 			Region:      aws.String(cfg.KmsRegion),
 			Endpoint:    aws.String(cfg.KmsEndpoint)},
 		)
+		// AWS uses IAM role for task
+		if cfg.BuildEnv == "production" {
+			sess, _ = session.NewSession(&aws.Config{
```
rupertbg, line 122: I'm pretty sure based on other AWS SDKs and the docs for the go SDK you can just put:
```sess, err := session.NewSession()```

The SDK will automatically seek the credentials it needs in env vars, instance metadata etc.

https://docs.aws.amazon.com/sdk-for-go/api/aws/session/

Comment (rupertbg) [reply go/batch-submitter/batch_submitter.go:122]: Nevermind, I see you are omitting creds in this case

File: go/batch-submitter/batch_submitter.go [hunk 3]
```diff
@@ -117,6 +117,13 @@ func Main(gitVersion string) func(ctx *cli.Context) error {
...
+				Region:   aws.String(cfg.KmsRegion),
+				Endpoint: aws.String(cfg.KmsEndpoint)},
```
rupertbg: Are these necessary in production? The endpoint should be default and the region should be inferred from the AWS_REGION or AWS_DEFAULT_REGION env var (if available)

boyuan-chen: We must pass the region. Otherwise, we will get `can not get public key from KMS for KeyId=20e255e9-8fdb-4f43-a0f3-d2520a81304f: MissingRegion: could not find region configuration`.

CAPtheorem (approved):

PR #141, opened 2022-04-30
Turing hello world

CAPtheorem:
Here is a simple SAM that sets up an endpoint for Turing with a working `curl` that allows people to emulate the type of data that will be POSTed by the Geth. Please check for correct function, and add to the documentation where needed.


wsdt (approved):

InoMurko:
cc @wsdt ideas with this?

wsdt:
@InoMurko as far as I remember this was a example for Turing. Haven't been involved in that, just reviewed but looked legitimate to me

InoMurko:
care to take another look @wsdt? and fix if things are broken. after you're happy let me know and I can review!

souradeep-das:
the contracts/ and tests/ seem to be duplicated here?

wsdt:
> care to take another look @wsdt? and fix if things are broken. after you're happy let me know and I can review!

sure!

wsdt:
Yeah seems to be obsolete, the same example also has been added to the `packages/boba/turing/` folder. 

The one in the packages folder seems to be more up-to-date since the test-file has been named correctly (the stableSwap is named hello-turing in this PR which is wrong). 

I think we can drop this PR.

PR #171, opened 2022-05-16
Wsdt/turing twitter pay

wsdt:
UI looks slightly different for Mainnet and testnest as AuthenticatedFaucet is only shown on Testnet. 

Everything seems to work fine, will have a look later or tomorrow for some final tests as that would become a more security-critical contract.


InoMurko:
@wsdt whats the status here?

wsdt:
Should actually be done. 

Would deploy it on Rinkeby first if approved, do some more tests and then go to mainnet since we transact real money here.

InoMurko:
@wsdt there's a confusion here.

Twitter pay seems to be `community`. 
But the changes are in boba_deployer and in the gateway.

Twitter pay should not be part of core infrastructure of Boba. It can be a separate thing the gateway only links to, but that is all.


Also, stuff in the `community` is not tested in our cicd. So we either drop this or find another way how to deal with this.

InoMurko:
Also, what is `turing-twitter` branch?

InoMurko:
If we want to have twitter pay as part of our core contracts it would be good to get an audit and move the code inside `packages/boba` and have it tested in GH actions + cirleci.

wsdt:
> @wsdt there's a confusion here.
> 
> Twitter pay seems to be `community`. But the changes are in boba_deployer and in the gateway.
> 
> Twitter pay should not be part of core infrastructure of Boba. It can be a separate thing the gateway only links to, but that is all.
> 
> Also, stuff in the `community` is not tested in our cicd. So we either drop this or find another way how to deal with this.

That was how it has been discussed at that time. The plan was to integrate it frontend-wise into the gateway. 

So moving the contracts itself to the core infra would fix the issue?

wsdt:
> Also, what is `turing-twitter` branch?

That is our current faucet (the Authenticated Twitter faucet)

wsdt:
> If we want to have twitter pay as part of our core contracts it would be good to get an audit and move the code inside `packages/boba` and have it tested in GH actions + cirleci.

will do!

wsdt:
done!

packages/boba/turing-twitter-pay/contracts/TwitterPay.sol
```diff
@@ -0,0 +1,92 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import "./interfaces/ITuringHelper.sol";
+import "@openzeppelin/contracts/utils/Strings.sol";
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+contract TwitterPay is Ownable {
```
InoMurko: @souradeep-das pls take a look at the contracts for twitter pay

souradeep-das: sure, just noticed, will get the review by tomorrow

boba_community/turing-hello-world/api/.aws-sam/build.toml
```diff
@@ -0,0 +1,13 @@
+# This file is auto generated by SAM CLI build command
+
```
souradeep-das: do we need this?

packages/boba/turing-twitter-pay/contracts/TwitterPay.sol
```diff
@@ -0,0 +1,92 @@
...
+        amountRegistersInLastEpoch = 0;
+        maxRegistersPerEpoch = maxRegistersPerEpoch_;
+    }
+
+    function setConfig(string memory apiUrl_, uint256 maxRegistersPerEpoch_) external onlyOwner {
+        apiUrl = apiUrl_;
+        maxRegistersPerEpoch = maxRegistersPerEpoch_;
+    }
+
+    function registerBobaBubble(string calldata twitterPostID_) external {
```
souradeep-das: noticed the "caller = wallet from bubbleId from tweet" is authenticated on the turing call. can we add a test for this case if we dont have one already 😄

packages/boba/turing-twitter-pay/contracts/TwitterPay.sol
```diff
@@ -0,0 +1,92 @@
...
+        require(isAllowedToReceive, string(abi.encodePacked("Invalid request:", Strings.toString(errorMsgVal))));
+
+        twitterRegister[authorId] = true;
+        bubbleRegister[bobaBubble] = _msgSender();
+
+        emit BubbleRegistered(authorId, bobaBubble);
+    }
+
+    function sendFundsMeta(address token_, address from_, uint256 bobaBubble_, uint256 amount_, bytes32 hashedMessage_, bytes memory signature_) external {
+        require(verifyMessage(hashedMessage_, signature_) == from_, "Signature faulty");
```
souradeep-das: I think the same pair (hashedMessage_, signature_) can be repeated for subsequent calls, irrespective of wether the signer signs any future messages/ metatxs?
and also since the `from_` doesn't have anything to do with this tx, is the meta tx redundant? (making the on-chain verification of signed message redundant)

just checked, a similar `verifyMessage` exists on the AuthenticatedFaucet too, should we change the signing logic or maybe do away with the necessity of user signing messages altogether?
(this is from what i understood)

packages/boba/turing-twitter-pay/contracts/TwitterPay.sol
```diff
@@ -0,0 +1,92 @@
...
+        _nonces[from_] = _nonces[from_] + 1;
```
souradeep-das: Is _nonces used somewhere? if it is, it could be tweaked by anyone as per will with a repeated pair((hashedMessage_, signature_) ?

packages/boba/turing-twitter-pay/contracts/TwitterPay.sol
wsdt: already have some :-)
- `should transfer funds to boba bubble (other)`
- `should transfer funds to boba bubble (yourself)`

packages/boba/turing-twitter-pay/contracts/TwitterPay.sol
wsdt: The meta tx is actually there to pay for user's gas. We do the transaction ourselves on behalf of the user. 
Or did I understand you wrong?

packages/boba/turing-twitter-pay/contracts/TwitterPay.sol
wsdt: Doesn't matter actually. No the nonce isn't used anywhere. It's actually just there to sign an arbitrary message and get the original sender that way since we do the transaction ourselves. We actually don't rely on a wallet in both examples but on the Twitter user (basically we don't care which wallet, just if the Twitter user is allowed to execute)

packages/boba/turing-twitter-pay/contracts/TwitterPay.sol
souradeep-das: I'll try to elaborate more, to make sure we are thinking on the same line (please do correct if im thinking wrong)

So my doubts are mostly around two things - i) if this qualifies to be a meta-transaction ii) do we need a meta-transaction for this operation

For i), i think a meta-transaction is in effect when a user signs a tx/message, and then the relayer comes in and submits it to the contract, and where in the contract the users signature is verified and the tx, or state change takes place. 
now going back to the `AuthenticatedFaucet` contract first, it totally makes sense that the user signs a message everytime, and the relayer then submits the message and signature to the contract to trigger sending funds. But, the way we have `verifyMessage` in the contract, the contracts allow the relayer to submit 'any' non-eip 712 (`hashedMessage`, signature) pair (irrespective of - if this signed message is the one user signed for faucet) and the contract would just pass verification  either way. And furthermore, the same (`hashedMessage`, signature) pair can be reused by the relayer later any number of times. Although, this is not a problem here, because the user is not getting deducted, but instead getting paid, but it breaks the essence of needing a signed message to trigger the faucet to send funds or a meta tx? (since the relayer can potentially just sendFunds without needing a message) maybe we could add a message verification mech on the contracts that allows every hashedMessage by the user only once, or maybe is a combination with their nonce

and in this contracts (TwitterPay) case - i think all of the above plus also, that the `verifyMessage()` just uses `from`, `hashedMessage` and `signature` which are all provided as inputs and doesn't feed into anything inward to the logic ? making this redundant or am I missing something? (in `AuthenticatedFaucet` in comparison, `to_` was used down to sendFunds)

which is also why i was thinking do we need users to sign messages at all? maybe we could just make the relayer call sendFunds() directly? (ii)
sorry this msg got too long 😓

packages/boba/turing-twitter-pay/contracts/TwitterPay.sol
souradeep-das: oh i see, so the hashedMessage = sha3(address, nonce_val_from_contract)
I think in that case, using a var nonce stored on the contract would make sense if we could reconstruct the hashedMessage on the contract instead of passing it as a param? and use this reconstructed to verify the signature?

packages/boba/turing-twitter-pay/contracts/TwitterPay.sol
```diff
@@ -0,0 +1,92 @@
...
+        sendFunds(token_, bobaBubble_, amount_);
+    }
+
+    /// @dev Send funds to authenticated user.
+    function sendFunds(address token_, uint256 bobaBubble_, uint256 amount_) public {
+        require(bubbleRegister[bobaBubble_] != address(0), "Unknown bubble");
+
+        IERC20(token_).transferFrom(_msgSender(), bubbleRegister[bobaBubble_], amount_);
```
souradeep-das: do we want the relayer to sends tokens from themselves?

packages/boba/turing-twitter-pay/contracts/TwitterPay.sol
wsdt: Oh yeah for the TwitterPay I actually didn't use the meta transaction (haven't even included it into the test). Just removed it, since I just let it in there in case we need it (as it was a fork from the AuthenticatedFaucet). 

But yeah, the idea (since it's very similar in the AuthenticatedFaucet) was, that we pay the gas for the user by simply verifying via signatures that the user actually owns this wallet (basically just signing an arbitrary message) + in combination with binding that wallet to a Twitter Account and Tweet about the BobaBubble (= part of a MD5 hash from the signing address).

souradeep-das: I see, great!
for the AuthenticatedFaucet- although its just rinkeby, so doesnt matter much- but just as a note/idea - one possible tweak could be to move the signature verification off-chain, since its independent to twitter verif any way - or if we could compute messageHash on-chain instead of params would make that fancy!

wsdt: Actually that was the main use case that we have a meta tx for the faucet, so that people can get testnet tokens without having ETH in their wallet (in contrary to the old faucet).

souradeep-das: thats true, but maybe people can just get testnet tokens after posting on twitter, without requiring any meta tx at all (like the ethereum faucets)
although if we chose to have meta tx, in that case, i think a selective hashedMessage and verifying mech would work better? Arbitrary message verification might not fit well as true meta txs

souradeep-das (approved):
:rocket:

InoMurko:
bump tests @wsdt

wsdt:
> bump tests @wsdt

starting today!

wsdt:
@InoMurko Tests pushed + included into GH actions. Will shortly add it to deployer.

packages/boba/gateway/src/services/networkService.js
```diff
@@ -576,6 +577,116 @@ class NetworkService {
...
 
+  async registerBobaBubble(tweetId) {
+    try {
+      const twitterPayContract = new ethers.Contract(
+        allAddresses.TwitterPay,
+        TwitterPayJson.abi,
+        this.L2Provider,
+      ).connect(this.provider.getSigner())
+
+      console.warn("HHH", this.provider.getSigner(), allAddresses.TwitterPay)
```
InoMurko: Is this a leftover?

packages/boba/gateway/src/services/networkService.js
```diff
@@ -576,6 +577,116 @@ class NetworkService {
...
+        )
+        return await transfer.wait()
+      }
+    } catch (error) {
+      return this.parseTwitterError(error)
+    }
+  }
+
+
+  parseTwitterError(error) {
```
InoMurko, line 648: Does this work well with Sentry? Cc @sk-enya 

For example, I see we're trying to rewrite certain errors - but I think a try catch is not the correct control flow here. Try catch catches unexpected errors too.

InoMurko:
@sk-enya if you could leave your opinion here as well please. It touches the gateway.

packages/boba/gateway/src/services/networkService.js
wsdt: Yes that has been taken into account. The original error message is retained, in case it cannot be mapped onto an expected error message.

wsdt:
@InoMurko added to deployer. 

I basically reverse-engineered. It seems to work, also rewrote the TS-tests for TwitterPay to test it with the local server, and the pre-deployed version to really ensure it works fine. 

Happy for feedback.

wsdt:
Referring to our last meeting, I would revert now the changes made that are related to the deployer and move it back to the `boba_examples/`. 

Would be great if one could confirm that before. @InoMurko

InoMurko:
Confirming @wsdt.
This is more of an example what can be built rather than something we'd try to build a business around it. At least not for now. So it should be in boba_examples.

wsdt:
> Confirming @wsdt. This is more of an example what can be built rather than something we'd try to build a business around it. At least not for now. So it should be in boba_examples.

thx! Should be done.

InoMurko:
this is merging into turing-twitter? why not develop?

wsdt:
> this is merging into turing-twitter? why not develop?

because the `twitter-pay` originates from the `twitter` (AuthenticatedFaucet) branch (and both haven't been merged at that time). Additionally, I made some minor improvements while implementing the TwitterPay on both projects to keep them consistent.

InoMurko (approved):
ship it

PR #172, opened 2022-05-18
Adjust exit fee in GPO service

boyuan-chen:
* Add subgraph for indexing events
* Modify exit fee based on the cost per message and price ratio

- [x] Add tests


packages/boba/gas-price-oracle/src/service.ts
```diff
@@ -824,6 +936,103 @@ export class GasPriceOracleService extends BaseService<GasPriceOracleOptions> {
     }
   }
 
+  private async _updateExitFee(): Promise<void> {
+    try {
```
InoMurko, line 940: this is a rather large function - would it be more readable if we broke it down?
also, can we start having wider code rows? :D

InoMurko:
@Boyuan-Chen why is this work needed? can you spend more time in filling in the requirements for this PR?
- Why is this needed?
- How is this fixed here?
- What is needed to deploy this and what are the effects to others/and other services?
And please put these into a GH PR template so that we always fill them out! Thanks

(and if your other PRs don't include the information above, fix them too please)

PR #175, opened 2022-05-26
Fix the problem of not being able to send all ETH

### LINKED ISSUE #174: eth_sendRawTransaction throws a error when account balances are equal cost
Hi there,
I wonder that when we create a transaction and send it with eth_sendRawTransaction RPC method, how the Geth can make sure the error(Insufficient funds for gas * price + value) was happened?

I mean i know the source code in here:

    // Transactor should have enough funds to cover the costs
// cost == V + GP * GL
if b := currentState.GetBalance(from); b.Cmp(tx.Cost()) < 0 {
	return core.ErrInsufficientFunds
}
my account balance: 0.004246
gas fee + value = 0.004246
I can be sure that my account balance is equal the cost, but it still throws "invalid transaction: insufficient funds for gas * price + value" error. why?

Thanks in advance!!

boyuan-chen:
`fee, err := fees.CalculateTotalFee(tx, s.RollupGpo)` in l2geth adds `l1SecurityFee` to fee, but we have already added it when we do `eth_estimateGas`. Therefore, fee is override by `gasPrice * gasLimit`
```
if isFeeTokenUpdate {
	fee = new(big.Int).Mul(tx.GasPrice(), new(big.Int).SetUint64(tx.Gas()))
}
```

Closed #174

### DIFF
--- a/integration-tests/test/fee-payment.spec.ts
+++ b/integration-tests/test/fee-payment.spec.ts
@@ -269,6 +269,65 @@ describe('Fee Payment Integration Tests', async () => {
     }
   })
 
+  it('{tag:other} should transfer all ETH with correct gas limit', async () => {
+    await setPrices(env, 1000)
+    const randomWallet = ethers.Wallet.createRandom().connect(env.l2Provider)
+
+    await env.l2Wallet.sendTransaction({
+      to: randomWallet.address,
+      value: ethers.utils.parseEther('1'),
+    })
+
+    // estimate gas fee
+    const estimatedGas = await randomWallet.estimateGas({
+      to: env.l2Wallet.address,
+      value: ethers.utils.parseEther('0.1'),
+    })
+    const gasPrice = await env.l2Provider.getGasPrice()
+    const estimatedGasFee = estimatedGas.mul(gasPrice)
+
+    // should transfer funds back
+    const balance = await randomWallet.getBalance()
+    await randomWallet.sendTransaction({
+      to: env.l2Wallet.address,
+      value: balance.sub(estimatedGasFee),
+      gasLimit: estimatedGas.toNumber(),
+    })
+
+    const postBalance = await randomWallet.getBalance()
+    expect(postBalance).to.be.eq(ethers.BigNumber.from('0'))
+
+    await env.l2Wallet.sendTransaction({
+      to: randomWallet.address,
+      value: ethers.utils.parseEther('1'),
+    })
+
+    // the gas fee should be constant
+    await randomWallet.sendTransaction({
+      to: env.l2Wallet.address,
+      value: balance.sub(estimatedGasFee).sub(BigNumber.from('10')),
+      gasLimit: estimatedGas.toNumber(),
+    })
+
+    const postBalance2 = await randomWallet.getBalance()
+    expect(postBalance2).to.be.eq(ethers.BigNumber.from('10'))
+
+    // should reject tx if gas limit is not provided
+    await env.l2Wallet.sendTransaction({
+      to: randomWallet.address,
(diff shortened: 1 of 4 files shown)


integration-tests/test/fee-payment.spec.ts
```diff
@@ -269,6 +269,65 @@ describe('Fee Payment Integration Tests', async () => {
     }
   })
 
+  it('{tag:other} should transfer all ETH with correct gas limit', async () => {
```
InoMurko, line 272: wouldn't this bug be much easier and quicker to test with a unit test?
image

InoMurko (changes requested):
just a comment about testing this - seems redundant to spawn the whole infrastructure to test this.

integration-tests/test/fee-payment.spec.ts
boyuan-chen: The unit tests are added, but the integration test is required, too. `verifyFee` only makes sure that the `gasLimit` is larger than `intrinsic gas` and the balance is enough to cover `gasPrice * gasLimit`, but it doesn't make sure that the tx can go through in `state_transition`.

InoMurko: why wouldn't it go through state transition? you didn't change the state transition in this PR.

boyuan-chen: There is a possibility that it fails at the `state_transaction` step. If we don't add this test in the integration test, then there is no way for us to test this case of transferring whole ETH from one account to another account.

boyuan-chen: Optimism doesn't need to consider this case, because it's very hard for their users to move all ETH from their accounts to another one. The l1 security fee is directly deducted from users' accounts, so basically it doesn't work in MetaMask or use a simple script.

If you insist on removing this test, I can delete them. However, we might not cover this case when we move to bedrock.

boyuan-chen: It's very hard to test this case directly in l2geth. The unit tests are only for one single function. The process of handling the transaction involves lots of different functions and that's why we need a final integration test to cover all cases.

InoMurko: I'm not against anything. I just want to understand.
What I see is that we push our testing way more into integration tests then unit tests because testing integration tests is easier. But we should be doing way more unit tests and cover all sorts of cases, and have less long running integration tests.

`There is a possibility that it fails at the state_transaction step.`
this I'm not sure why - because if you're verifying fee here, additional checks would need to be performed in state transition?

boyuan-chen: `VerifyFee` only makes sure that `gasLimit` is larger than `intrinsic gas`. `intrinsic gas` is not what final gas is consumed (gasUsed). This involves a situation that users add `gasLimit` in their payload which is lower than the actual estimated gas limit. This tx can go through this `verifyFee`, but it fails at `state_transaction`.

InoMurko: Ahaaaa. Makes sense now!

InoMurko (approved):

integration-tests/test/fee-payment.spec.ts
boyuan-chen: Another case that might fail at the `state_transaction`. If `eth_estimateGas` returns the wrong value and users pass in the payload, the tx will fail (out of gas). The `intrinsic gas` for transferring ETH is `21000`, so any `tx.Gas()` is larger than `21000` can pass this `verifyFee`. They might fail at `state_transaction` if `gasLimit` is lower than `21000 + l1SecurityFee / l2GasPrice`.

PR #184, opened 2022-06-07
Calldata compression

### LINKED ISSUE #180: calldata compression
> https://github.com/Boyuan-Chen/calldata-compression/tree/main/result

We can save about **28.49%** by compressing calldata that goes into the `CTC` contract using `Zlib`.
We can save about **42.82%** by compressing calldata that goes into the `CTC` contract using `Brotli`.

The compression and gas saving results for the last 100K single transactions:
 
| Method          | Compression Ratio | Estimated Fee Savings |
| --------------- | ----------------- | ----------------- |
| Zlib            | 24.34%            | -8.15% |
| Zlib Dictionary | 22.36%            | -11.12% |
| Brotli          | 25.03%            | -5.48% |

I studied 5K events. The compression and gas saving results for the batch transaction that goes into the `CTC` contract:

| Method          | Compression Ratio | Estimated Fee Savings |
| --------------- | ----------------- | ----------------- | 
| Zlib            | 51.08%            | 29.47% | 
| Zlib Dictionary | 51.05%           | 29.44% |
| Brotli          | 59.93%           | 42.96% |

I also did the speed test for decompressing the data. Since only `data transport layer` needs to decompress the payload, I ran the speed test in `node.js`.

| Method          | Number of Batches | Time Duration |
| --------------- | ----------------- | ----------------- | 
| Brotli            | 5385           | 8.2058988s | 
| Zlib          | 5385           | 3.022597383 |

The average l2 block number in one batch is 82. The worst case and best case for decompressing data is 

| Method          | Number of Batches | Time Duration |
| --------------- | ----------------- | ----------------- | 
| Zlib (best case)           | 100           | 1.6ms | 
| Brotli (best case)         | 100          | 32.2ms |
| Zlib (worst case)           | 100           | 42.4ms | 
| Brotli (worst case)         | 100          | 166.8ms |

> The worst case means that each batch has 272 L2 blocks and the best case means that each batch only has 2 L2 blocks. I found those two cases and got the rawTransactions. I copied and pasted those data 100 times and used them as the input for the above calculation.

boyuan-chen:
> Calldata compression discussion #180 - closed #180

The calldata is compressed in the `batch submitter`. The `core-utils` handles the decompression. The `data transport layer` uses functions from `core-utils` and stores the decoded payloads in the database and shares with `l2geth`.

Unit tests are added and one new integration test is added. The new integration test uses the same method in `data transport layer` to decompress the calldata and search the transactions in `sequencer` and `verifier` nodes. This test helps us know that the `verifier` and `sequencer` have the same blocks and sync up correctly.


go/batch-submitter/drivers/sequencer/encoding.go
```diff
@@ -105,30 +188,74 @@ type AppendSequencerBatchParams struct {
...
+	case BatchTypeLegacy:
+		// Write each length-prefixed tx.
+		for _, tx := range p.Txs {
+			_ = writeUint64(w, uint64(tx.Size()), TxLenSize)
+			_, _ = w.Write(tx.RawTx()) // can't fail for bytes.Buffer
+		}
+	case BatchTypeBrotli:
+		bw := brotli.NewWriterLevel(w, 11)
+		for _, tx := range p.Txs {
+			if err := writeUint64(bw, uint64(tx.Size()), TxLenSize); err != nil {
```
InoMurko: Why do we have more error handling here than above?

boyuan-chen: `writeUint64()` writes `TxLenSize` to bw. In other words, it compress TxLenSize first. It might fail if `brotli` can't compress the payload.

InoMurko: why wouldn't brotli compress bytes?

InoMurko: I wanna double check because I see `writeUint64` was purposely adapted to now return an error, it previously ignored one

go/batch-submitter/flags/flags.go
```diff
@@ -196,6 +196,12 @@ var (
 		Value:  1,
 		EnvVar: prefixEnvVar("BLOCK_OFFSET"),
 	}
+	SequencerBatchType = cli.StringFlag{
```
InoMurko: is this needed? don't we go brotli only?

InoMurko (approved):

go/batch-submitter/flags/flags.go
boyuan-chen: We can disable the calldata compression if we want to.

go/batch-submitter/drivers/sequencer/encoding.go
boyuan-chen: Ideally, this won't fail. Since we don't write the `Brotli` package, it's safer to check it. If it can't be compressed and we ignore the error, then the payload is totally wrong and the CTC contract is messed up.

go/batch-submitter/flags/flags.go
InoMurko: I don't want to nitpick but ... why would we disable compression once it was live?
I'm just thinking in terms of dead code.

boyuan-chen: Ok! Will remove!

InoMurko:
Awesome!

packages/data-transport-layer/test/unit-tests/services/l1-ingestion/handlers/sequencer-batch-appended.spec.ts
```diff
@@ -1,11 +1,21 @@
 import { BigNumber, ethers } from 'ethers'
+import { sequencerBatch, add0x, BatchType } from '@eth-optimism/core-utils'
+
+const compressBatchWithZlib = (calldata: string): string => {
```
InoMurko: zlib leftovers

packages/data-transport-layer/test/unit-tests/services/l1-ingestion/handlers/sequencer-batch-appended.spec.ts
```diff
@@ -58,5 +68,54 @@ describe('Event Handlers: CanonicalTransactionChain.SequencerBatchAppended', ()
...
+
+    describe('mainnet transactions', () => {
+      for (const mock of mockData) {
+        const { input, output } = mock
+        const { event, extraData, l2ChainId } = input
+        const hash = mock.input.extraData.l1TransactionHash
+
+        it(`uncompressed: ${hash}`, () => {
+          // Set the type to be legacy
+          output.transactionBatchEntry.type = BatchType[BatchType.LEGACY]
```
InoMurko, line 80: am I wrong at assuming that batch type legacy is applicable only if we want to keep the ability to NOT compress batches?

afaik - the moment we deploy - all new batches are compressed, the only reason why we would have "legacy" batches is for the verification to be able to read the uncompressed batches  - like (/core-utils/src/optimism/batch-encoding.ts for DTL). correct?

go/batch-submitter/drivers/sequencer/encoding.go
```diff
@@ -71,6 +76,29 @@ func (c *BatchContext) Read(r io.Reader) error {
...
 }
 
+// BatchType represents the type of batch being submitted. When the first
+// context in the batch has a timestamp of 0, the blocknumber is interpreted as
+// an enum that represets the type.
+type BatchType int8
+
+const (
+	// BatchTypeLegacy represets the legacy batch type.
+	BatchTypeLegacy BatchType = -1
```
InoMurko, line 86: I don't think this is needed anymore, right?
If all batches after deploy are compressed batches - brotli - why have different types of batches.
Or am I missing something in the implementation?

packages/data-transport-layer/test/unit-tests/services/l1-ingestion/handlers/sequencer-batch-appended.spec.ts
boyuan-chen: Yes

go/batch-submitter/drivers/sequencer/encoding.go
boyuan-chen: It's used for the Reader function. The Reader function is only used for the unit tests. We have to test the case of uncompressed data to the original data.

InoMurko:
```
We have to test the case of uncompressed data to the original data.
```
in batch submitter? why?

boyuan-chen: This is the test data.
```
    {
      "name": "complex",
      "hex_encoding": "0102030405060708000004000102030405060708090a0b0c0d0e0f000102030405060708090a0b0c0d0e0f000102030405060708090a0b0c0d0e0f000102030405060708090a0b0c0d0e0f00000ac980808080808080808000000ac980808080808080808000000ac980808080808080808000000ac9808080808080808080",
      "should_start_at_element": 4328719365,
      "total_elements_to_append": 395016,
      "contexts": [
        {
          "num_sequenced_txs": 258,
          "num_subsequent_queue_txs": 197637,
          "timestamp": 25887770890,
          "block_number": 47446822415
        },
        {
          "num_sequenced_txs": 258,
          "num_subsequent_queue_txs": 197637,
          "timestamp": 25887770890,
          "block_number": 47446822415
        },
        {
          "num_sequenced_txs": 258,
          "num_subsequent_queue_txs": 197637,
          "timestamp": 25887770890,
          "block_number": 47446822415
        },
        {
          "num_sequenced_txs": 258,
          "num_subsequent_queue_txs": 197637,
          "timestamp": 25887770890,
          "block_number": 47446822415
        }
      ],
      "txs": [
        "c9808080808080808080",
        "c9808080808080808080",
        "c9808080808080808080",
        "c9808080808080808080"
      ],
      "error": false
    },
```


The actual tx data is decoded (Reader) from `hex_encoding` and `contexts`. The batch submitter compresses data using the `Write()` function and decompresses data using `Reader()` function. We need to compare the actual tx data with the decompressed data from `Reader()` function. It's very hard to compare the tx data without the `BatchTypeLegacy ` type in the Reader function that actually decodes the payload.

PR #204, opened 2022-06-20
Turing Multi-Value

mmontour1306:
Relax the length restriction to support larger Turing payloads. Also removes the length field from the off-chain interaction, so that the off-chain endpoint deals directly with ABI-encoded bytes. Overloads the internal "rType" field with a version number to preserve legacy semantics for existing contracts.

Before merging, we need to decide where to set the limit on payload size. We need to check that we don't expose any attack scenarios where the system operator is forced to spend ETH storing very large amounts of calldata into the L1 contracts (i.e. making sure that the individual user is charged a correct amount of gas, even for failed transactions).


packages/boba/turing/test/002_local_math.ts
```diff
@@ -0,0 +1,252 @@
...
+
+if (hre.network.name === "boba_local") {
+  describe("Basic Math", function () {
+
+      before(async () => {
+
+      var http = require('http')
+      var ip = require("ip")
+
+      var server = module.exports = http.createServer(async function (req, res) {
```
InoMurko: @wsdt remember our conversation around having a local webserver (instead of remote lambda/serverless)

InoMurko:
This increase of payload size probably significantly (please correct me) reduces the amount of transaction items in a batch. But how much?

souradeep-das:
> We need to check that we don't expose any attack scenarios where the system operator is forced to spend ETH storing very large amounts of calldata

Also, is there a way we could charge the L1 sec fee dynamically - depending on the calldata size, although we are taking the gas upfront. maybe asking for extra gas with gas-refunds?

packages/boba/turing/test/002_local_math.ts
wsdt: great example thx!

InoMurko:
Leaving here so that we remember to change the docs: https://docs.boba.network/turing/turing#important-properties-of-turing

wsdt:
@InoMurko I would also add "keep warm" for the lambda services, since for most Turing use cases it's not acceptable for the first transaction to fail. Already having the SAM template in our new project (SBT)

InoMurko:
Absolutely!

mmontour1306:
I've pushed some pending changes but this current branch is a dead end. I'll be starting from a clean copy of 'devel' and then manually re-applying changes rather than trying to resolve merge conflicts through git.

wsdt:
@mmontour1306 ok thx! Would be awesome if you then just tag me on the new branch to not run against the old one

mmontour1306:
I've pushed a forced update of my current branch, re-applying the changes on top of a more recent Develop. It is not ready for merging yet.

boba_examples/turing-hello-world/contracts/TuringHelper.sol
```diff
@@ -114,6 +116,22 @@ contract TuringHelper is ITuringHelper, Ownable {
      need to include a timestamp and/or more details about the
      offchain interaction.
   */
+  function TuringTxV1(string memory _url, bytes memory _payload)
```
wsdt: so V1 will stay the new one then? Wouldn't it make sense to name it V2 maybe?

InoMurko: good point :-)

InoMurko (approved):
LGTM!

wsdt (approved):

boyuan-chen:
Can we update all examples in `boba_examples` and `boba_community`?

souradeep-das (approved):
LGTM 🚀

boyuan-chen:
> Can we update all examples in `boba_examples` and `boba_community`?

Ah, never mind. `TuringTx` will still work right?

boyuan-chen (approved):
Looks good to me. BTW, can we update the docs?

l2geth/internal/ethapi/api.go
```diff
@@ -915,6 +915,27 @@ func DoCall(ctx context.Context, b Backend, args CallArgs, blockNrOrHash rpc.Blo
...
+		return nil, 0, false, err
+	}
+	p2, err := b.SuggestL2GasPrice(ctx)
+	if err != nil {
+		log.Error("TURING Can't get L2 gas price")
+		return nil, 0, false, err
+	}
+
+	if p2.BitLen() > 0 {
+		evm.Context.TuringGasMul = float64(p1.Uint64()) / float64(p2.Uint64())
```
boyuan-chen, line 931: Is it compatible with the old turing tx? We require more gas usage for turingTx, which means that old turingTx might be failure for the verifier and replica.

wsdt:
Just tested on our university project with complex types etc. 
-> works - will merge now as discussed.

PR #230, opened 2022-07-02
Split by timings for parallel run

### LINKED ISSUE #193: unreliable integration tests
Integration tests should not be coupled with each other (such that we need to preserve order of how they're executed) because it breaks test isolation and parallelism.

Integration tests are also very time consuming and we should utilize `--split-by=timings`. https://circleci.com/blog/a-guide-to-test-splitting/

Notifying @souradeep-das @Boyuan-Chen

trangtran-enya:
[:clipboard: Add associated issues, tickets, docs URL here.](https://github.com/bobanetwork/boba/issues/193)

## Overview

Currently, we already execute testcases in parallel. However, testcases are divided based on filename instead of execution time. 
[Issue 139](https://github.com/bobanetwork/boba/issues/193)

## Changes

- Add `mocha-junit-reporter` to collect timing data
- Turn off `hardhat-gas-reporter` as it conflicts with `mocha-junit-reporter`. Need discussion?
- Turn on split by timings
- split the time it takes for all tests to run in half

## Testing

- integration tests


ops/up_local.sh
```diff
@@ -35,30 +35,30 @@ if [[ $BUILD == 1 ]]; then
...
-    docker-compose build -- replica
+  docker-compose build --parallel --no-cache -- builder l2geth l1_chain
+  docker-compose build --parallel --no-cache -- deployer dtl batch_submitter relayer integration_tests
+  docker-compose build --no-cache -- boba_message-relayer-fast
+  docker-compose build --no-cache -- gas_oracle
+  docker-compose build --no-cache -- boba_deployer
+  docker-compose build --no-cache -- fraud-detector
+  docker-compose build --no-cache -- monitor
+  docker-compose build --no-cache -- verifier
+  docker-compose build --no-cache -- replica
```
InoMurko: how come caches are disabled?

integration-tests/hardhat.config.ts
```diff
@@ -5,7 +5,8 @@ import '@nomiclabs/hardhat-ethers'
 import '@nomiclabs/hardhat-waffle'
 import 'hardhat-gas-reporter'
 
-const enableGasReport = !!process.env.ENABLE_GAS_REPORT
+// const enableGasReport = !!process.env.ENABLE_GAS_REPORT
+const enableGasReport = false
```
InoMurko: why doesn't this work?

ops/up_local.sh
trangtran-enya: Testcases in the beginning failed, so I tried to run with image built from latest code.

integration-tests/hardhat.config.ts
trangtran-enya: Looks like the result from GasReport overwrite Timing result, which is used for future run.

ops/up_local.sh
trangtran-enya: In the beginning, this worked. All testcases passed. Now, some testcases using cross-chain message failed.

integration-tests/hardhat.config.ts
InoMurko: Have you tried things like:
https://github.com/mochajs/mocha/pull/1360
https://www.npmjs.com/package/mocha-multi-reporters
?

trangtran-enya: No, not yet. Let me try. Thanks Ino

The GasReporter is this one, configured through Hard-Hat. Maybe we will meet some problem. But I think I can figure out.

souradeep-das (approved):
lgtm

integration-tests/test/boba-fee-payment.spec.ts
```diff
@@ -1044,9 +1044,9 @@ describe('Boba Fee Payment Integration Tests', async () => {
         bobaBalance
       )
       await transferBackTx.wait()
-    })
+    }).retries(3)
```
wsdt, line 1047: why do we have a retry here? Just retry three times until we don't have enough Boba tokens anymore?

wsdt (approved):
looks great! 🔥

integration-tests/test/boba-api.spec.ts
```diff
@@ -68,8 +68,8 @@ describe('Boba API Tests', async () => {
       process.env.L2_BOBA_ADDRESS = L2Boba.address
     })
 
-    describe('Mainnet', async () => {
-      it('{tag:boba} should swap BOBA for ETH', async () => {
```
boyuan-chen, line 72: Do we want to remove all tags?

InoMurko: eventually 🗡️

integration-tests/test/boba-fee-payment.spec.ts
InoMurko: flaky test or flaky infra!

InoMurko (approved):

Issue #231, opened 2022-07-04
Boba deployer regularly hangs integration tests


InoMurko:
## Issue Type

```
[ x] bug report
[ ] feature request
```

## Current Behavior
```
boba_deployer_1              | {
[...]

PR #232, opened 2022-07-04
Inomurko/231 boba deployer fix

InoMurko:
Closes (https://github.com/bobanetwork/boba/issues/231)

## Overview

Occasionally the boba deployer would die because of hiccups in DAO deployment. 

## Changes

- Grep docker compose logs and detect DAO exit and restart the boba deploy container.

## Testing

/ Should reduce occasional integration test failure.


boyuan-chen (approved):

boyuan-chen:
There is one problem when we redeploy the`boba-deployer`. The `CDM Fast` address is changed and it's different from what we have in SDK. The message relayer fast won't work.

InoMurko:
> There is one problem when we redeploy the`boba-deployer`. The `CDM Fast` address is changed and it's different from what we have in SDK. The message relayer fast won't work.

Can you tell me more? How is it that the CDM Fast is registering a stale address?

boyuan-chen:
> > There is one problem when we redeploy the`boba-deployer`. The `CDM Fast` address is changed and it's different from what we have in SDK. The message relayer fast won't work.
> 
> Can you tell me more? How is it that the CDM Fast is registering a stale address?

The CDM Fast address is hardcoded in https://github.com/bobanetwork/boba/blob/6e034ed2817105846a19aac3f50241fade1e77f1/packages/sdk/src/utils/contracts.ts#L127. Redeploying the CDM Fast contract will generate another contract address. The message relayer fast still uses the contract address in SDK, because it uses SDK to query and analyze events.

InoMurko:
> > > There is one problem when we redeploy the`boba-deployer`. The `CDM Fast` address is changed and it's different from what we have in SDK. The message relayer fast won't work.
> 
> > 
> 
> > Can you tell me more? How is it that the CDM Fast is registering a stale address?
> 
> 
> 
> The CDM Fast address is hardcoded in https://github.com/bobanetwork/boba/blob/6e034ed2817105846a19aac3f50241fade1e77f1/packages/sdk/src/utils/contracts.ts#L127. Redeploying the CDM Fast contract will generate another contract address. The message relayer fast still uses the contract address in SDK, because it uses SDK to query and analyze events.

Aha. Yes, makes sense!

packages/boba/contracts/deploy/013-BobaDao.deploy.ts
```diff
@@ -175,24 +172,14 @@ const deployFn: DeployFunction = async (hre) => {
...
-
-  await setPendingAdminTx.wait()
+  let eta1
+  let setPendingAdminData
+  try {
+    ;[eta1, setPendingAdminData] = await setPendingAdmin(hre, eta_delay_s)
+  } catch (error) {
+    console.log(`setPendingAdmin failed because of : ${error}`)
+    ;[eta1, setPendingAdminData] = await setPendingAdmin(hre, eta_delay_s)
+  }
```
InoMurko, line 182: @souradeep-das from your experience, does a try-catch catch a failed deployment transaction?
I've had 10 successful runs in cirleci integration tests.it seems so?

souradeep-das: oh yes, a failed tx should go right into the catch block
great, this looks like a smart way to do this for the time being!! 😎

souradeep-das (approved):

wsdt (approved):
looks great!

PR #237, opened 2022-07-07
enable bobascan on gateway

InoMurko:
## Overview
This fixes Brotli integration into the gateway (monorepo issues)
This fixes Bobascan urls for all chains.

## Changes

- Brotli is implemented in the native ZLIB
- Urls in gateway
- ...

### DIFF
--- a/packages/boba/gateway/src/util/masterConfig.js
+++ b/packages/boba/gateway/src/util/masterConfig.js
@@ -78,8 +78,8 @@ if (process.env.REACT_APP_CHAIN === 'rinkeby') {
         chainId: 288,
         chainIdHex: '0x120',
         rpcUrl: `https://mainnet.boba.network`,
-        blockExplorer: `http://bobascan.com/`,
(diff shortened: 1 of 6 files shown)


packages/boba/gateway/src/util/masterConfig.js
```diff
@@ -78,8 +78,8 @@ if (process.env.REACT_APP_CHAIN === 'rinkeby') {
         chainId: 288,
         chainIdHex: '0x120',
         rpcUrl: `https://mainnet.boba.network`,
-        blockExplorer: `https://blockexplorer.boba.network/`,
-        transaction: `https://blockexplorer.boba.network/tx/`,
+        blockExplorer: `http://bobascan.com/`,
```
boyuan-chen: Can we use `https://` instead of `http://`

InoMurko: Uuuups!

PR #257, opened 2022-07-19
add veBoba contracts

souradeep-das:
:clipboard: ref: moving https://github.com/bobanetwork/ve-boba/pull/13

## Overview

Adds all veBoba contracts and tests according to design spec https://github.com/bobanetwork/ve-boba/issues/4

## Changes

 - Adds the following contracts:
   - ve (token lock for ve)
   - gauge/gaugeFactory (vault/gauge pools to direct emissions towards)
   - voter (voting and allocating voting weight to gauges)
   - dispatcher (Boba treasury that allocates weekly emissions)
 - Adds deploy scripts and tests

## Testing

set up env and
```
yarn test
```


packages/boba/ve-boba/.env.example
```diff
@@ -0,0 +1,3 @@
+L2_NODE_WEB3_URL=http://localhost:8545
+DEPLOYER_PRIVATE_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
```
wsdt, line 2: Is this key used somewhere? Just the local one right?

packages/boba/ve-boba/contracts/BaseV1-dispatcher.sol
```diff
@@ -0,0 +1,113 @@
...
+    function notifyRewardAmount(uint amount) external;
+}
+
+// this is the layer-2 Boba custodian contract which holds and distributes emission to the voter
+
+contract BaseV1Dispatcher is Ownable {
+
+    uint internal constant week = 86400 * 7; // allows minting once per week (reset every Thursday 00:00 UTC)
+    uint internal weeklyEmission;
+    underlying public immutable _token;
```
wsdt, line 35: was confused for a sec. Is this a forked contract or self-written? 
If we wrote it ourselves, shouldn't we maybe follow general code styleguides such as `IUnderlying`, or at least `Underlying`. 

Super minor remark. same for others

packages/boba/ve-boba/contracts/BaseV1-dispatcher.sol
```diff
@@ -0,0 +1,113 @@
...
+        _token = underlying(ve(__ve).token());
+        _voter = voter(__voter);
+        _ve = ve(__ve);
+        active_period = (block.timestamp + (2*week)) / week * week;
+    }
+
+    // load up the custodian contract with initAmount, which should cover for sum amounts
+    // if we want initial ve dist, to be a percent owner, make sure
+    // sum amounts / initAmount = % ownership of top protocols
+    function initialize(
```
wsdt, line 61: Using `initializer` and `initialize` let's me always think about proxies. **Super minor** again, but if it is self-written/not forked, then maybe let's use a different name for better maintainability/readability?

packages/boba/ve-boba/contracts/BaseV1-voter.sol
```diff
@@ -0,0 +1,331 @@
...
+interface IBaseV1GaugeFactory {
+    function createGauge(address, address, address) external returns (address);
+}
+
+interface IGauge {
+    function notifyRewardAmount(address token, uint amount) external;
+    function getReward(address account, address token) external;
+}
+
+interface IMinter {
```
wsdt, line 39: E.g. here we follow common style guides, and in the same file above with `ve` etc not.

packages/boba/ve-boba/contracts/BaseV1-voter.sol
```diff
@@ -0,0 +1,331 @@
...
+        _vote(_tokenId, _poolVote, _weights);
+    }
+
+    function _vote(uint _tokenId, address[] memory _poolVote, int256[] memory _weights) internal {
+        _reset(_tokenId);
+        uint _poolCnt = _poolVote.length;
+        int256 _weight = int256(ve(_ve).balanceOfNFT(_tokenId));
+        int256 _totalVoteWeight = 0;
+        int256 _totalWeight = 0;
+        int256 _usedWeight = 0;
```
wsdt, line 162: Everything super minor today sorry, but wouldn't it save some gas if we don't assign default values? Since we are on L2 that really doesn't matter, just wanted to say it.

packages/boba/ve-boba/contracts/Basic-gauges.sol
```diff
@@ -0,0 +1,113 @@
...
+    constructor(address _pool, address  __ve, address _voter, address _operator) {
+        pool = _pool;
+        _ve = __ve;
+        voter = _voter;
+        transferOwnership(_operator);
+    }
+
+    // simple re-entrancy check
+    uint internal _unlocked = 1;
+    modifier lock() {
```
wsdt, line 56: I think I remember know, these contracts were forks right? Since I remember writing about the `ReentrancyGuard` already in a previous PR with the same/similar code. Because then it's not worth the hassle to adapt these things for sure.

wsdt (approved):
Looks good to me, I think I reviewed this code already somewhere else right? 

Just left some super minor remarks, can be ignored but might make sense maybe if it's easy to refactor (and is not forked from somewhere else, because then I always tend to stick to the original).

packages/boba/ve-boba/.env.example
souradeep-das: yep the local one!

packages/boba/ve-boba/contracts/BaseV1-dispatcher.sol
souradeep-das: (min diff with solidly BaseV1-minter.sol)

packages/boba/ve-boba/contracts/BaseV1-dispatcher.sol
souradeep-das: (min diff with solidly BaseV1-minter.sol)

packages/boba/ve-boba/contracts/BaseV1-voter.sol
souradeep-das: (min diff with solidly BaseV1-voter.sol)

souradeep-das:
Hey @wsdt, yes you have reviewed this before. Just moving all of that to the monorepo :) 
These are excellent points and totally make sense.
I have tried to retain as much structure from Solidly as was possible - in effect to making the diff shorter, and easier review. And that has crept in some of their choices too, looks like the three points above fall in that. So maintaining the same then, without changes?

wsdt:
Yes, minimizing the diff is much more important! thank you

InoMurko (approved):

PR #270, opened 2022-07-22
feat: modify community DAO to work with veBoba

souradeep-das:
:clipboard: https://github.com/bobanetwork/ve-boba/projects/1#card-82399384

## Overview

Updates the community DAO to work with veNFTs from veBoba as the voting power (only).
By removing the usage of Boba( and xBoba) in its base form, and extending the application of veNFTs to the DAO, voting power across the ecosystem is unanimous.

## Changes

 - allow to specify veNFTs to vote with/ propose with
 - get voting power at a prior timestamp instead of a prior block-number
 - remove threshold requirements for cancelling proposal
 - changed voting proposal thresholds and quorumVotes amount
 - removed unused variables (we are going for a fresh depl instead of a proxy update)

Current broader-mechanisms of proposing, queueing, executing is still the same including the use of - voting delay, voting period, and counting votes from a common checkpoint for a proposal.

## Testing

integration-tests/test/pool_dao_actions.spec.ts


packages/boba/contracts/contracts/DAO/governance/GovernorBravoDelegate.sol
```diff
@@ -225,70 +247,77 @@ contract GovernorBravoDelegate is GovernorBravoDelegateStorageV1, GovernorBravoE
...
-        // the first voter in the voting period also sets the voting power snapshot
-        if (proposal.startBlock == 0) {
-          proposal.startBlock = sub256(block.number, 1);
+        uint256 totalVePower;
+        for (uint i = 0; i < tokenIds.length; i++) {
+          // check if token is not already used on this proposal
+          Receipt storage receipt = veReceipts[proposalId][tokenIds[i]];
+          require(receipt.hasVoted == false, "GovernorBravo::castVoteInternal: tokenId already used for proposal");
+          // check if owner or approved
+          // ve locks that have been merged with other locks are burned, and cannot be used here
```
souradeep-das, line 294: if someone merges their veNFTs, token A + token B -> token B
tokenA would be burnt and they would not be able to use tokenA to vote on proposals that are on their "voting period" right now. (token B would not have the cumulative ve power because, we are checking the power at a prior timestamp)

On the gateway, If we have a merge screen on the ve locks - we should add a warning, something like "You might not be able to use the voting power of the first token, on proposals that are 'currently on their voting period'. We suggest you to vote(if any) before continuing with the merge"
cc @sk-enya

packages/boba/contracts/contracts/DAO/governance/GovernorBravoDelegate.sol
```diff
@@ -45,23 +45,26 @@ contract GovernorBravoDelegate is GovernorBravoDelegateStorageV1, GovernorBravoE
       * @param timelock_ The address of the Timelock
       * @param bobaToken_ The address of the L2 BOBA token
       * @param xbobaToken_ The address of the L2 xBOBA token
+      * @param ve_ The address of the ve lock contract
       * @param votingPeriod_ The initial voting period in seconds
       * @param votingDelay_ The initial voting delay in seconds
       * @param proposalThreshold_ The initial proposal threshold
       */
-    function initialize(address timelock_, address bobaToken_, address xbobaToken_, uint votingPeriod_, uint votingDelay_, uint proposalThreshold_) public {
+    function initialize(address timelock_, address bobaToken_, address xbobaToken_, address ve_, uint votingPeriod_, uint votingDelay_, uint proposalThreshold_) public {
```
souradeep-das: the contracts fit the pattern to be proxy updated, 
except adding the `ve` address needs to be determined from this
https://github.com/bobanetwork/boba/discussions/276

packages/boba/contracts/contracts/DAO/governance/GovernorBravoDelegate.sol
InoMurko: Excellent point, but for making sure it’s not lost, create an issue.

InoMurko (commented):
Is xboba kept in for backwards compatibility?

souradeep-das:
> Is xboba kept in for backwards compatibility?

I was thinking, If we want to maintain and update with the proxy, its safest to keep the inherited slot orders same and only add. Both `boba` and `xBoba` shouldn't be needed anymore, (we could possibly replace only one of them with the ve address though- deleting both changes the order, so to be safe have maintained the order)
We could remove all of the unused variables, if we chose to do a fresh depl

packages/boba/contracts/contracts/DAO/governance/GovernorBravoDelegate.sol
souradeep-das: oh yes thanks!
https://github.com/bobanetwork/boba/issues/278

souradeep-das:
ref Discussion: https://github.com/bobanetwork/boba/discussions/276
If the decision is to redeploy a fresh DAO, I will remove the unused vars (boba, xBoba) from here

boyuan-chen (approved):

wsdt (approved):
awesome!

PR #272, opened 2022-07-22
Wsdt/reference optimistic geth

wsdt:

## Overview

Add reference-optimistic-geth to submodules. 

## Changes

- Reference-optimistic-geth submodule added. 
- Merge from develop

## Testing

GH actions


.gitmodules
```diff
@@ -1,3 +1,11 @@
...
 	path = l2geth/tests/testdata
 	url = https://github.com/ethereum/tests
+
+[submodule "bedrock/erigon"]
+	path = bedrock/erigon
+	url = https://github.com/bobanetwork/erigon
+
+[submodule "bedrock/reference-optimistic-geth"]
+	path = bedrock/reference-optimistic-geth
+	url = https://github.com/bobanetwork/reference-optimistic-geth
```
InoMurko, line 11: what do we plan to do with geth?

wsdt: you said we should include it as well not?

InoMurko: Oh did I? 😅
We can always remove it - maybe it makes sense to have it. I guess you and @mmontour1306 decide how useful it is.

wsdt: ok at least I understood it that way, sorry for that! Well here is the new PR, if we want to merge it great otherwise fine as well :)
https://github.com/bobanetwork/boba/pull/274

This branch had a weird state

PR #309, opened 2022-08-19
Add bobabeam deployment

boyuan-chen:

## Overview

Add bobabeam deployment files and fix some issues

## Changes

- Add bobabeam deployment files
- Fix monitor for getting the onramp txs
- Add price ratio in the gas oracle service. It fetch the token price from `Coingecko API` and update the market price ratio and price ratio.

## Testing

Add unit tests for gas price oracle service


packages/boba/gas-price-oracle/src/service.ts
```diff
@@ -633,4 +656,55 @@ export class GasPriceOracleService extends BaseService<GasPriceOracleOptions> {
...
+    try {
+      const URL = `https:///api.coingecko.com/api/v3/coins/${id}?localization=false&tickers=false&community_date=false&developer_data=false&sparkline=false`
+      const payload = await fetch(URL)
+      const payloadParsed = await payload.json()
+      return Number(payloadParsed.market_data.current_price.usd)
+    } catch (err) {
+      this.logger.warn(`CAN\'T QUERY TOKEN PRICE ${err} - ${id}`)
+      return 0
+    }
+  }
```
InoMurko: it would be very good if we used at least one more data provider.
1. if coingeko goes down
2. if coingeko has skewed info
3. weighted average of multiple providers

boyuan-chen: The second source is added.

packages/boba/gas-price-oracle/src/service.ts
```diff
@@ -659,10 +665,17 @@ export class GasPriceOracleService extends BaseService<GasPriceOracleOptions> {
...
-      const l1NativeTokenPrice = await this._getTokenPrice(
-        this.options.l1TokenId
-      )
+      /* eslint-disable */
+      const BobaPriceFromCoinGecko = await this._getTokenPriceFromCoinGecko('boba-network')
+      const l1NativeTokenPriceFromCoinGecko = await this._getTokenPriceFromCoinGecko(this.options.l1TokenCoinGeckoId)
+      const BobaPriceFromCoinMarketCap = await this._getTokenPriceFromCoinMarketCap('14556')
+      const l1NativeTokenPriceFromCoinMarketCap = await this._getTokenPriceFromCoinMarketCap(this.options.l1TokenCoinMarketCapId)
+
+      // calculate the average price of the two sources
```
InoMurko, line 679: my hunch what could happen is: an outlier!

lets imagine the real price of boba is 10$. 
coinmarketcap reports 10$
but geko reports 9899$, because their rounding is faulty (9.899$)

it means that the average is still wrong. Wdyt we should do about it?
it feels like having more sources to get the price, and remove outliers is the way to go.

boyuan-chen: Add the function of filtering the outliers and fetch the data of exchanges

packages/boba/gas-price-oracle/src/service.ts
```diff
@@ -670,10 +670,16 @@ export class GasPriceOracleService extends BaseService<GasPriceOracleOptions> {
       const l1NativeTokenPriceFromCoinGecko = await this._getTokenPriceFromCoinGecko(this.options.l1TokenCoinGeckoId)
       const BobaPriceFromCoinMarketCap = await this._getTokenPriceFromCoinMarketCap('14556')
       const l1NativeTokenPriceFromCoinMarketCap = await this._getTokenPriceFromCoinMarketCap(this.options.l1TokenCoinMarketCapId)
+      const BobaMarketPricesFromCoinMarketCap = await this._getTokenMarketPriceFromCoinMarketCap('14556')
```
InoMurko, line 673: should this be a constant variable? is this basically `const CoinMarketCapBobaId`?

boyuan-chen: No, CoinMarketCap has exchange prices. That function returns the exchange prices of BOBA token.

packages/boba/gas-price-oracle/test/gas-price-oracle.spec.ts
```diff
@@ -647,6 +647,12 @@ describe('gas-price-oracle', () => {
 
     await tempGasPriceOracleService.init()
 
+    const input = [-100, 2, 2, 2, 2, 2, 2, 2, 2, 10000]
+    const filteredOutliers = tempGasPriceOracleService.filterOutliers(input)
+    expect(filteredOutliers.includes(10000)).to.be.eq(false)
+    expect(filteredOutliers.includes(-100)).to.be.eq(false)
+    expect(filteredOutliers.length).to.be.eq(input.length - 2)
+
```
InoMurko: this should really fit into a separate test :)

boyuan-chen: Fixed

InoMurko: and cases more cases tested?

how often do you gather prices?
this only partially addresses the problem of misbehaving API.
you could have skewed prices for an outage that lasts for hours, then all your tilting would go towards the outlier since you're calculating the weights from past prices, no?

boyuan-chen: We update prices every 10 minutes.

I am lost here. Do you want me to find another two or three APIs to get prices?

PR #328, opened 2022-08-29
Redirect the Hybrid Compute requests through a proxy service.

mmontour1306:

## Overview

This routes hybrid compute requests through a new proxy service, allowing those requests to be firewalled away from any internal network destinations. This implementation uses an unmodified SOCKS5 docker image which I found, but we may want to use a different protocol or develop a custom proxy with additional features (such as adding logging and analytics for the requests).

## Changes

 Changes to be committed:
	modified:   core/vm/evm.go
	modified:   rpc/client.go
	modified:   ../ops/docker-compose.yml

## Testing

The existing Turing tests should still run and should be able to access the same AWS endpoints when this change is applied to a local system. Other test environments will likely need additional changes to support the new Docker service.


boyuan-chen:
Can we put env in `l2geth/cmd/utils/flags.go` and import that in `evm.go`?

mmontour1306:
Yes, it should be possible to add the proxy string to l2geth/cmd/utils/flags.go.

InoMurko:
After Bedrock.

PR #329, opened 2022-08-30
fix: Lzbridge max transfer limit

souradeep-das:
:clipboard: 

## Overview

Enables the "max Transfer Amount per Day" limit for all cases

## Changes
- added the 'check' for all cases 
- did not refactor

## Testing

yarn test


boyuan-chen (approved):
YES! I forgot that one.

InoMurko (approved):

PR #358, opened 2022-09-16
fixes for proposal and lock record fixes

sk-enya:

## Overview
Fixes for annoying console errors from fetchProposal, Lock Records
## Changes

- Describe your changes and implementation choices. More details make PRs easier to review.
- Fixes for annoying console errors from fetchProposal, Lock Records
- Imported the beamer image in ecosystem.

## Testing

Open gateway and check consoles.

### DIFF
--- a/packages/boba/gateway/src/services/networkService.js
+++ b/packages/boba/gateway/src/services/networkService.js
@@ -227,8 +227,6 @@ class NetworkService {
       this.networkGateway
     ).post('/', { jsonrpc: "2.0", method: "status", id: 1 })
 
-    console.log("Verifier response: ", response)
-
     if (response.status === 200) {
       const status = response.data.result
       return status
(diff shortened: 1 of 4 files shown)


packages/boba/gateway/src/services/networkService.js
```diff
@@ -4829,7 +4829,13 @@ class NetworkService {
       return
     }
 
+    if (!+process.env.REACT_APP_ENABLE_LOCK_PAGE) {
```
boyuan-chen, line 4825: Is it a new environment variable?

boyuan-chen (approved):

packages/boba/gateway/src/services/networkService.js
```diff
@@ -4152,17 +4150,20 @@ class NetworkService {
...
         let endTimestamp = proposalData.endTimestamp.toString()
 
         let proposal = await delegateCheck.getActions(i+2)
 
         let hasVoted = null
 
-        if( this.account ) {
-          hasVoted = await delegateCheck.getReceipt(proposalID, this.account)
-        }
+        // NOTE: as per the veBoba getReciept expecting params as proposalId & tokenId so this will break
```
souradeep-das: should we remove this then?

souradeep-das (approved):

packages/boba/gateway/src/services/networkService.js
sk-enya: yeah as we have Lock Page so to disable that we need to have this.

Default value should be 0.

Whenever we have to enable the lock page we will remove this param.

packages/boba/gateway/src/services/networkService.js
sk-enya: Yeah, okay we can remove this.

PR #407, opened 2022-10-10
Add Teleportation contract and service

boyuan-chen:

## Overview

We introduce the teleportation service for users to bridge BOBA tokens between L2s.

When users deposit BOBA tokens to our teleportation contract, our teleportation service can detect the tx and release the funds on other chains. We also include serval security protocols to protect users and us.

## Changes

- The new teleportation contract is added and deployed to these `Proxy__Teleportation: 0xd68809330075C792C171C450B983F4D18128e9BF ` and `Teleportation: 0x502310DADE5BAB1065c788F6b98B78f5fc49FC09 `.
- The teleportation service is added. It monitors all BOBA L2s and release the funds when a new deposit is founded on other chains.
- The docker file is added and the teleportation service is added to `docker-compose-side.yml`

## Related PR

#448 - Add contract addresses to the registered files for Alt L2s
#449 - Teleportation docs

## Testing

The unit tests and integration tests are added


packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,503 @@
...
+     * @param _maxDepositAmount The initial maximum deposit amount.
+     */
+    function initialize(
+        address _BobaTokenAddress,
+        uint256 _minDepositAmount,
+        uint256 _maxDepositAmount
+    )
+        external
+        onlyOwner()
+        onlyNotInitialized()
```
wsdt, line 175: we could remove that here since `initializer()` already checks for initialization. 
https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/proxy/utils/Initializable.sol

Boyuan-Chen: Removed!

boyuan-chen: We must add it. Otherwise, it's reverted. 

Reason:

```
    bool public _initialized;

    /**
     * @dev Indicates that the contract is in the process of being initialized.
     */
    bool public _initializing;

    /**
     * @dev Modifier to protect an initializer function from being invoked twice.
     */
    modifier initializer() {
        require(_initializing || !_initialized, "Initializable: contract is already initialized");
   }

The `initializer()` reverted when `_initializing` and `_initialized` are `false`.
```

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,518 @@
+// SPDX-License-Identifier: MIT
```
InoMurko, line 1: what is the status of this contract in terms of audits?

boyuan-chen: We can review it internally first. It takes serval days to add unit tests and integration tests, so it's better to send it for the audit before finishing everything.

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,516 @@
...
+        uint256 indexed chainId,
+        bool supported
+    );
+
+    /**********************
+     * Function Modifiers *
+     **********************/
+
+    modifier onlyDisburser() {
+        require(msg.sender == disburser || disburser == address(0), 'Caller is not the disburser');
```
souradeep-das: should we move away from "disburser == address(0)" pattern?
(like the newer contracts)

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,516 @@
...
+        _;
+    }
+
+    modifier onlyOwner() {
+        require(msg.sender == owner || owner == address(0), 'Caller is not the owner');
```
souradeep-das: should we move away from "owner == address(0)" pattern?

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,516 @@
...
+        _;
+    }
+
+    modifier onlyAltL2s() {
+        require(address(BobaTokenAddress) == 0x4200000000000000000000000000000000000006, "Only alt L2s can call this function");
+        _;
+    }
+
+    modifier onlyNotAltL2s() {
+        require(address(BobaTokenAddress) != 0x4200000000000000000000000000000000000006, "Only not alt L2s can call this function");
```
souradeep-das: how about "only non alt L2s"?

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,516 @@
...
+     * @param _amount The amount of BOBA to deposit.
+     * @param _toChainId The destination chain ID.
+     */
+    function teleportNativeBOBA(uint256 _amount, uint256 _toChainId)
+        external
+        payable
+        onlyAltL2s()
+        whenNotPaused()
+    {
+        require(_amount >= minDepositAmount, "Deposit amount is too small");
```
souradeep-das: should we just set '_amount = msg.value' instead of asking it as a param? to reduce cost?

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,516 @@
...
+        for (uint256 i = 0; i < _numDisbursements; i++) {
+            uint256 _amount = _disbursements[i].amount;
+            address _addr = _disbursements[i].addr;
+            uint256 _sourceChainId = _disbursements[i].sourceChainId;
+            uint256 _depositId = _disbursements[i].depositId;
+
+            // Ensure the depositId matches our expected value.
+            require(_depositId == totalDisbursements[_sourceChainId], "Unexpected next deposit id");
+            unchecked {
+                totalDisbursements[_sourceChainId] += 1;
```
souradeep-das: do we want to make this unchecked though? since we use it above and its not limited

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,516 @@
...
+            }
+
+            // Deliver the dispursement amount to the receiver. If the
```
souradeep-das:
```suggestion
            // Deliver the disbursement amount to the receiver. If the
```

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,516 @@
...
+            // disbursement fails, the amount will be kept by the contract
+            // rather than reverting to prevent blocking progress on other
+            // disbursements.
+
+            // slither-disable-next-line calls-loop,reentrancy-events
+            (bool success, ) = _addr.call{ value: _amount, gas: 1000000 }("");
```
souradeep-das: too much gas for native token transfer?

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,516 @@
...
+        for (uint256 i = 0; i < _numDisbursements; i++) {
+            uint256 _amount = _disbursements[i].amount;
+            address _addr = _disbursements[i].addr;
+            uint256 _sourceChainId = _disbursements[i].sourceChainId;
+            uint256 _depositId = _disbursements[i].depositId;
+
+            // Ensure the depositId matches our expected value.
+            require(_depositId == totalDisbursements[_sourceChainId], "Unexpected next deposit id");
+            unchecked {
+                totalDisbursements[_sourceChainId] += 1;
```
souradeep-das: same as above

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,516 @@
...
+            }
+
+            // Deliver the dispursement amount to the receiver. If the
+            // disbursement fails, the amount will be kept by the contract
+            // rather than reverting to prevent blocking progress on other
+            // disbursements.
+
+            // slither-disable-next-line calls-loop,reentrancy-events
+            try IERC20(BobaTokenAddress).transfer(_addr, _amount) {
```
souradeep-das: this should never fail? should we allow this to revert instead, if this ever does?

also should we use 'safeTransfer' for consistency

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,516 @@
...
+    /**
+     * @dev transfer disburser role to new address
+     *
+     * @param _newDisburser new disburser of this contract
+     */
+    function transferDisburser(
+        address _newDisburser
+    )
+        external
+        onlyDisburser()
```
souradeep-das: minor: not sure how we want to do this- 
onlyOwner() or onlyDisburser() ?

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,516 @@
...
+    }
+
+
+    /**
+     * @notice Sets the minimum amount that can be deposited in a receive.
+     *
+     * @param _minDepositAmount The new minimum deposit amount.
+     */
+    function setMinAmount(uint256 _minDepositAmount) external onlyOwner() {
+        emit MinDepositAmountSet(minDepositAmount, _minDepositAmount);
```
souradeep-das: lets store the old value in a new var, and emit the event at the end?

applies for `setMaxAmount()` and `setMaxTransferAmountPerDay()` too

packages/boba/contracts/test/endToEndTests/teleportation.spec.ts
```diff
@@ -0,0 +1,780 @@
...
+          depositId: 2,
+        },
+      ]
+      await L2Boba.approve(Proxy__Teleportation.address, _amount)
+      await expect(Proxy__Teleportation.disburseBOBA(payload))
+        .to.emit(Proxy__Teleportation, 'DisbursementSuccess')
+        .withArgs(2, signerAddress, _amount, 4)
+    })
+
+    it('should not disburse BOBA tokens if the depositId is worng', async () => {
```
souradeep-das: minor
```suggestion
    it('should not disburse BOBA tokens if the depositId is wrong', async () => {
```

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,516 @@
...
+        disburser = msg.sender;
+        owner = msg.sender;
+
+       // set maximum amount of tokens can be transferred in 24 hours
+        transferTimestampCheckPoint = block.timestamp;
+        maxTransferAmountPerDay = 100_000e18;
+
+        __Context_init_unchained();
+        __Pausable_init_unchained();
+        __ReentrancyGuard_init_unchained();
```
souradeep-das: reentrancy guard is unused as of now, any plans to use it?

packages/boba/contracts/contracts/Teleportation.sol
boyuan-chen: Ideally, it will never fail. `safeTransfer` is the internal function, so we can't wrap it into try{}catch{}.

Anyway, I will remove try{}catch{}.

packages/boba/contracts/contracts/Teleportation.sol
boyuan-chen: I don't think we have to do that, since we don't fire txs based on the events.

packages/boba/contracts/contracts/Teleportation.sol
boyuan-chen: Remove!

packages/boba/contracts/contracts/Teleportation.sol
souradeep-das: true, but for convention and consistency, or maybe if someone else listens to the events?

boyuan-chen: Fixed!

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,503 @@
...
+import '@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol';
+import "@openzeppelin/contracts-upgradeable/security/PausableUpgradeable.sol";
+
+/**
+ * @title Teleportation
+ *
+ * Shout out to optimisim for providing the inspiration for this contract:
+ * https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts/contracts/L1/teleportr/TeleportrDeposit.sol
+ * https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts/contracts/L2/teleportr/TeleportrDisburser.sol
+ */
```
InoMurko, line 19: if this contract is a mesh of Optimisms contracts, which I presume were audited, the better way of adding this contract is to:
1. add them unchanged
2. apply changes

this way we'd know from git what the diff is

wsdt (approved):
Looks good to me 🤔

packages/boba/teleportation/README.md
```diff
@@ -0,0 +1,21 @@
...
+## Configuration
+
+All configuration is done via environment variables. See all variables at [.env.example](.env.example); copy into a `.env` file before running.
+
+|                             | Description                             | Default |
+| --------------------------- | --------------------------------------- | ------- |
+| L2_NODE_WEB3_URL            | The endpoint of Layer 2                 |         |
+| TELEPORTATION_DISBURSER_KEY | The pk of disburser                     |         |
+| POLLING_INTERVAL            | The polling interval of fetching events | 60s     |
+| EVENT_PER_POLLING_INTERVAL  | The polling range of events             | 1000    |
```
InoMurko: what does this mean? number of blocks you're looking at for events?

yeah - this needs a better name. BLOCK_RANGE or similar

packages/boba/teleportation/src/exec/run.ts
```diff
@@ -0,0 +1,88 @@
...
+  config.load({
+    env: true,
+    argv: true,
+  })
+
+  const env = process.env
+  const L2_NODE_WEB3_URL = config.str('l2-node-web3-url', env.L2_NODE_WEB3_URL)
+  // This private key is used to send funds to the contract and initiate the tx,
+  // so it should have enough BOBA balance
+  const TELEPORTATION_DISBURSER_KEY = config.str(
```
InoMurko, line 28: this is the private key that controls boba on source and destination chain, right?

packages/boba/teleportation/src/service.ts
```diff
@@ -0,0 +1,332 @@
...
+        startBlock,
+        endBlock
+      )
+      events = [...events, ...partialEvents]
+      startBlock = endBlock
+    }
+    return events
+  }
+
+  async _putDepositInfo(
```
InoMurko, line 298: whats the purpose of storing this on the filesystem?

packages/boba/teleportation/src/exec/run.ts
```diff
@@ -0,0 +1,88 @@
...
+  const service = new TeleportationService({
+    l2RpcProvider: l2Provider,
+    chainId,
+    teleportationAddress: TELEPORTATION_ADDRESS,
+    bobaTokenAddress: BOBA_TOKEN_ADDRESS,
+    disburserWallet,
+    selectedBobaChains,
+    pollingInterval: POLLING_INTERVAL,
+    eventPerPollingInterval: EVENT_PER_POLLING_INTERVAL,
+    dbPath: DATABASE_PATH,
```
InoMurko, line 83: is this a filesystem database? why?

packages/boba/teleportation/src/service.ts
```diff
@@ -0,0 +1,332 @@
...
+        )} - latestBlock: ${latestBlock}`
+      )
+      this._putDepositInfo(chainId, latestBlock)
+    } catch (e) {
+      this.logger.error(e)
+    }
+  }
+
+  // get events from the contract
+  async _getEvents(
```
InoMurko, line 274: a number of functions don't have unit tests

InoMurko (changes requested):

packages/boba/teleportation/src/exec/run.ts
boyuan-chen: It's the disburser key, not admin key.

InoMurko:
do we need finality protection debth (ie reorg)?

packages/boba/teleportation/src/service.ts
boyuan-chen: We store the last synced block, so it don't have to re-sync from the starting block, like the data transport layer.

packages/boba/teleportation/src/service.ts
boyuan-chen: No, it's tested. `_getEvents` is nested in `_watchTeleportation`. The `_watchTeleportation` is tested.

packages/boba/teleportation/src/exec/run.ts
boyuan-chen: It's used for storing the last synchronized block

packages/boba/teleportation/src/service.ts
InoMurko: DTL uses LevelDB, correct?

boyuan-chen:
> do we need finality protection debth (ie reorg)?

We don't have reorg on L2, but I can add the block confirmation. That doesn't make anything different.

packages/boba/teleportation/src/exec/run.ts
InoMurko: whats the difference?
in this scenario it's the private key that unlocks boba that's sent from A to B.

InoMurko: it's important that it hase enough liquidity

packages/boba/teleportation/src/service.ts
boyuan-chen: Yes, it's the same purpose.

packages/boba/teleportation/src/exec/run.ts
boyuan-chen: The balance will be monitored in the monitor service. Disburser has some BOBA tokens and can disburse the BOBA token, but it doesn't control the contract (pause, unpause, parameters). 

The worst case of leaking the disburser key is that we lose all BOBA tokens in the disburser key, but BOBA tokens in the smart contract are safe. Only admin can withdraw BOBA token from the smart contract.

souradeep-das (approved):
new changes LGTM!

packages/boba/teleportation/README.md
boyuan-chen: Updated the name to `BLOCK_RANGE_PER_POLLING`.

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -250,7 +251,7 @@ contract Teleportation is PausableUpgradeable {
 
         IERC20(BobaTokenAddress).safeTransferFrom(msg.sender, address(this), _amount);
 
-        emit BobaReceived(_toChainId, totalDeposits[_toChainId], msg.sender, _amount);
+        emit BobaReceived(_toChainId, block.chainid, totalDeposits[_toChainId], msg.sender, _amount);
```
InoMurko: the naming of chainIds could be done better
what if we used `from` and `to`?
```
emit BobaReceived(_toChainId, block.chainid, totalDeposits[_toChainId], msg.sender, _amount);
```
teleport boba to `_toChainId ` from `block.chainid`, which increments nonce (index of totalDeposits, from the sender, for amount.

Is that right?

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,510 @@
...
+            require(supportedChains[_sourceChainId], "Source chain is not supported");
+            totalDisbursements[_sourceChainId] += 1;
+
+            // Deliver the disbursement amount to the receiver. If the
+            // disbursement fails, the amount will be kept by the contract
+            // rather than reverting to prevent blocking progress on other
+            // disbursements.
+
+            // slither-disable-next-line calls-loop,reentrancy-events
+            (bool success, ) = _addr.call{ value: _amount }("");
```
InoMurko: you don't have to specify gas?

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -0,0 +1,510 @@
...
+        for (uint256 i = 0; i < _numDisbursements; i++) {
+            uint256 _amount = _disbursements[i].amount;
+            address _addr = _disbursements[i].addr;
+            uint256 _sourceChainId = _disbursements[i].sourceChainId;
+            uint256 _depositId = _disbursements[i].depositId;
+
+            // Ensure the depositId matches our expected value.
+            require(_depositId == totalDisbursements[_sourceChainId], "Unexpected next deposit id");
+            require(supportedChains[_sourceChainId], "Source chain is not supported");
+            totalDisbursements[_sourceChainId] += 1;
```
InoMurko, line 332: unchecked?

packages/boba/teleportation/src/exec/run.ts
InoMurko: does this private key always use BOBA to pay fees or will it need to hold ETH on boba eth?

packages/boba/teleportation/src/exec/run.ts
InoMurko: would using a managed postgres be better for performance and reliability?

packages/boba/teleportation/src/service.ts
InoMurko: the idea behind a unit tests is to test the smallest, still testable unit - will it be a function or a method.
in _watchTeleportation, you're not testing only getEvents, but also 
```
try {
       lastBlock = await this._getDepositInfo(chainId)
     } catch (e) {
       this.logger.warn(`No deposit info found in chainId - ${chainId}`)
       lastBlock = depositTeleportation.height
       // store the new deposit info
       await this._putDepositInfo(chainId, lastBlock)
     }
```
So you're interested that `_getEvents` meets its design and behaves as intended.
- Does _getEvents really check all blocks that you've set from and to?
- Can you make sure that `_getEvents` stays within this.options.blockRangePerPolling?
- Can you make sure that `_getEvents` appends all events if endBlock needs chunking?
- Can you make sure that `_getEvents` includes all events from the contract queryFilter?

InoMurko: same applies for `_putDepositInfo` that I think would benefit from more gradual testing

packages/boba/contracts/contracts/Teleportation.sol
boyuan-chen: Yes

packages/boba/contracts/contracts/Teleportation.sol
boyuan-chen: Yup!

packages/boba/contracts/contracts/Teleportation.sol
boyuan-chen: @souradeep-das 's comment for this `do we want to make this unchecked though? since we use it above and its not limited`, so I removed the unchecked. It costs a little more gas.

packages/boba/contracts/contracts/Teleportation.sol
InoMurko: I want to learn why! :-D

boyuan-chen: I don't know exactly why, but I found this. https://github.com/ethereum/solidity/releases/tag/v0.6.4

souradeep-das: ahaa, nice catch! lets specify some gas here so the receiver does not get a lot of gas to play with? (_addr can be anything)

boyuan-chen: Ok!

InoMurko: I caught something 👌🏻🥇😂

packages/boba/teleportation/src/exec/run.ts
boyuan-chen: I don't think that's necessary. Even we lose the data, it only takes about 10 seconds to sync up. We can also manually add it if we want to.

packages/boba/contracts/contracts/Teleportation.sol
boyuan-chen: Fixed

packages/boba/contracts/contracts/Teleportation.sol
boyuan-chen: Added

packages/boba/teleportation/src/exec/run.ts
boyuan-chen: We can also increase the `eventPerPollingInterval` to speed up.

packages/boba/contracts/contracts/Teleportation.sol
```diff
@@ -337,7 +337,7 @@ contract Teleportation is PausableUpgradeable {
             // disbursements.
 
             // slither-disable-next-line calls-loop,reentrancy-events
-            (bool success, ) = _addr.call{ value: _amount }("");
+            (bool success, ) = _addr.call{ gas: 21000, value: _amount }("");
```
souradeep-das: great, does 21000 increase the gas ask by too much initially?
if it actually does, a normal transfer should use 2300, a little more than that should be good? (we can update though)

boyuan-chen: Updated to 3000.

packages/boba/teleportation/src/exec/run.ts
InoMurko: hmmm... if it takes 10 seconds to sync. why do it at all? wouldn't it be cleaner and simpler if you just dropped database and have a stateless service?

packages/boba/teleportation/src/service.ts
```diff
@@ -0,0 +1,332 @@
...
+      let sliceStart = 0
+      let sliceEnd = numberOfDisbursement > 10 ? 10 : numberOfDisbursement
+      while (sliceStart < numberOfDisbursement) {
+        const slicedDisbursement = disbursement.slice(sliceStart, sliceEnd)
+        const totalDisbursements = slicedDisbursement.reduce((acc, cur) => {
+          return acc.add(BigNumber.from(cur.amount))
+        }, BigNumber.from('0'))
+        if (
+          this.options.bobaTokenAddress !==
+          '0x4200000000000000000000000000000000000006'
```
InoMurko: should this be pulled out into a constant?

packages/boba/teleportation/src/service.ts
```diff
@@ -0,0 +1,332 @@
...
+      }
+
+      // sort disbursement
+      disbursement = orderBy(disbursement, ['depositId'], ['asc'])
+      // disbure the token
+      await this._disburseTx(disbursement, chainId, latestBlock)
+    }
+  }
+
+  async _disburseTx(
```
InoMurko, line 225: does this have a test?

packages/boba/teleportation/src/service.ts
```diff
@@ -0,0 +1,332 @@
...
+    const events = await this._getEvents(
+      depositTeleportation.Teleportation,
+      this.state.Teleportation.filters.BobaReceived(),
+      lastBlock,
+      latestBlock
+    )
+    return events
+  }
+
+  async _disbureTeleportation(
```
InoMurko:
```suggestion
  async _disburseTeleportation(
```
also, does this have a test?

packages/boba/teleportation/src/exec/run.ts
boyuan-chen: Well currently it only takes about 1 or 2 seconds to sync. I mean it probably takes 1 minutes to sync in the future if we have 10M blocks. We can improve this in the future if we meet some problems.

packages/boba/teleportation/src/service.ts
boyuan-chen: Fixed

packages/boba/teleportation/src/service.ts
boyuan-chen: Added

packages/boba/teleportation/src/service.ts
boyuan-chen: Fixed

InoMurko (approved):
send it!

Issue #404, opened 2022-10-06
[Gateway] How to handle the special case of ERC20 tokens in gateway?


boyuan-chen:
## Issue Type

```
[ ] bug report
[x] feature request
```

## Current Behavior

https://github.com/bobanetwork/boba/pull/403

We should have a standard way to handle the special tokens that are not ERC20 compatible in our gateway.

## Steps to Reproduce

1.
2.
3.

sk-enya:
@Boyuan-Chen can you please help how do we have to handle the special tokens. can you please add any example tokens and it's way of handling?

sk-enya:
looks already completed!

PR #408, opened 2022-10-11
A global way to get token info in gateway

boyuan-chen:
:clipboard: Add associated issues, tickets, docs URL here. https://github.com/bobanetwork/boba/issues/404

## Overview

Some tokens are not ERC20 compatible, which means that our standard ABI doesn't work for those tokens.

In order to solve the problem, we store token information in one separate file. In our gateway, it can load token information instead of sending PRC calls. This should improve the speed too.

## Changes

- Add token information files
- Load name, symbol and decimal from the information file

## Testing


packages/boba/gateway/src/services/networkService.js
```diff
@@ -142,7 +146,8 @@ if (APP_CHAIN === 'rinkeby') {
     L1LPAddress: addresses_Mainnet.Proxy__L1LiquidityPool,
     L2LPAddress: addresses_Mainnet.Proxy__L2LiquidityPool,
     ...bobaBridges,
-    ...l0Protocols
+    ...l0Protocols,
+    tokenInfo: tokenInfoMainnet,
```
sk-enya: Instead of putting this in allAddress we should use separate var for tokenInfo.

sk-enya (approved):

packages/boba/register/addresses/tokenInfoMainnet.json
```diff
@@ -0,0 +1,276 @@
+{
```
InoMurko: if you intend to use this on alt-l1s - perhaps it makes sense to name the files by chainid?
<chainid>-mainnet.json
<chainid>-testnet.json

boyuan-chen: For alt-l1s, I will use `tokenInfo[L1_NAME][MAINNET or TESTNET]`. It's hard to remember 10 chain Ids.

packages/boba/gateway/src/services/networkService.js
boyuan-chen: Fixed!

packages/boba/register/addresses/tokenInfoMainnet.json
InoMurko: well... you can make a mapping of chains, chainids as well. Better this way than having different implementations across the stack

InoMurko: and you get the chainid with rpc and immediately know what to load

boyuan-chen: You mean adding all token information with chain Ids in the `tokenInfoMainnet` file? Like this:
```
{
  chain_id_1: {},
  chain_id_2: {}
}
```

boyuan-chen: > and you get the chainid with rpc and immediately know what to load

The loading process is before connecting to the RPC endpoint. I need to double check if you want to automatically load the file.

InoMurko: yeah that would work too, tokenInfo.json as chainid doesn't tell you if it's testnet or not

Boyuan-Chen: One for mainnet and one for testnet. That will be simpler and it should be global for alt l1s except adding token information to the file.

boyuan-chen: Ok. Now, we only need to add token information to `tokenInfo.json` for alt l1s. The gateway automatically fetches the right information based on the chain Id.

InoMurko (approved):
🙌

PR #461, opened 2022-11-17
V2 Random Numbers for hybrid compute

mmontour1306:

## Overview

Adds a new protocol for generating secure random numbers across multiple transactions. The client and server each
supply randomness in a commit/reveal scheme which ensures that neither side can influence the result. The
implementation includes a server-side state cache (currently in-memory only) which as a side effect means that the
random result will not change between an eth_estimateGas() call and the actual transaction.

This PR is filed to allow people to look at the code, but it is not ready to be merged yet. Among other issues, the
server logs currently leak the random numbers prior to the reveal phase. The in-memory cache should be stored persistently on the local disk so that transaction sequences will not be disrupted by a node restart. Documentation needs to be updated.

## Testing

The boba_examples/turing-hello-world local tests have been updated to include the new functionality as well as the previous mode.


mmontour1306:
Unless there's specific customer demand for this in our current system, I'm leaning toward bumping this feature into a future V3 (Anchorage/Erigon) release.

mmontour1306:
Closing this PR. Feature will be considered for Anchorage/V3.

PR #466, opened 2022-11-21
remove boba turing credit predeplox for proxy

InoMurko:
Confusing naming for boba credit proxy and removed reference to predeploy 0x4200000000000000000000000000000000000020 which is the implementation of boba credit.


souradeep-das (approved):
aah great!

boyuan-chen (approved):

PR #477, opened 2022-11-23
min alt l2 token fee adjustment

InoMurko:
## Overview

Ability to update the minimum secondary fee token amount a user needs to hold.
## Changes

- add `updateSecondaryFeeTokenMinimum` method to `Boba_GasPriceOracle`
- 

## Testing

Addiing tests to `Boba_GasPriceOracle.spec.ts`

### DIFF
--- a/packages/contracts/contracts/L2/predeploys/Boba_GasPriceOracle.sol
+++ b/packages/contracts/contracts/L2/predeploys/Boba_GasPriceOracle.sol
@@ -89,7 +89,7 @@ contract Boba_GasPriceOracle {
     event UpdateDecimals(address, uint256);
     event WithdrawBOBA(address, address);
     event WithdrawSecondaryFeeToken(address, address);
-    event UpdateSecondaryFeeTokenMinimum(address, uint256);
+    event UpdateSecondaryFeeTokenMinimum(uint256, uint256);
 
     /**********************
      * Function Modifiers *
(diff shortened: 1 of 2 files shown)


packages/contracts/contracts/L2/predeploys/Boba_GasPriceOracle.sol
```diff
@@ -69,6 +69,9 @@ contract Boba_GasPriceOracle {
     // Decimals of the price ratio
     uint256 public decimals = 0;
 
+    //Minimum alt token
+    uint256 public secondaryFeeTokenMinimum = 2e15;
```
souradeep-das, line 73: minor: since we moved this under a proxy, i think we could avoid all the direct var inits for the contract 😄

packages/contracts/contracts/L2/predeploys/Boba_GasPriceOracle.sol
```diff
@@ -320,6 +325,17 @@ contract Boba_GasPriceOracle {
...
 
+    /**
+     * Update the minimum secondary fee token minimum
+     * @param _secondaryFeeTokenMinimum the minimum amount
+     */
+    function updateSecondaryFeeTokenMinimum(uint256 _secondaryFeeTokenMinimum) public onlyOwner {
+        // Users should have more than 0.002 l1 native token
+        require(_secondaryFeeTokenMinimum >= 2e15);
+        secondaryFeeTokenMinimum = _secondaryFeeTokenMinimum;
+        emit UpdateSecondaryFeeTokenMinimum(owner(), _secondaryFeeTokenMinimum);
```
souradeep-das: minor: would emitting (oldValue, newValue) as event be better?

souradeep-das (approved):
🚀

packages/contracts/contracts/L2/predeploys/Boba_GasPriceOracle.sol
InoMurko: yes!

boyuan-chen (approved):

Issue #504, opened 2022-12-22
[ AA ] bundler needs to be added to docker-compose-side.yml


InoMurko:
## Issue Type

```
[ ] bug report
[ ] feature request
```

## Expected Behavior
account abstraction is put into a docker container - boba_deployer_aa  with the bundler contracts! (purge them from bundler) - if it’s too tied/coupled - lets recheck our assumptions
account abstraction also deploys samples, entrypoint contract is registered on address manager
bundler fetches the entrypoint contract from address manager

souradeep-das:
closed by #531 and #508

InoMurko:
thank you for cleaning up @souradeep-das

PR #509, opened 2023-01-05
run and build bundler in the side docker-compose

InoMurko:
[:clipboard: Add associated issues, tickets, docs URL here.](https://github.com/bobanetwork/boba/issues/504)

## Overview
Shows how to build Bundler service and loads it in the docker compose side. Integrates our Address Manager to retrieve deployed contracts done in a separate pr from @souradeep-das  

## Changes

- Alpine image with two layers
- Import address manager from contracts and use it to retrieve two registered contracts if address manager is present

## Testing

Will add more tests once things are tied together with @souradeep-das


ops/docker-compose-side.yml
```diff
@@ -69,3 +69,33 @@ services:
...
+    restart: on-failure
+    environment:
+      url: https://${HOST}/
+      LETSENCRYPT_HOST: $HOST
+      LETSENCRYPT_PORT: 3000
+      VIRTUAL_HOST: $HOST
+      VIRTUAL_PATH: /
+      VIRTUAL_DEST: /
+      MIN_BALANCE: 0
+      MNEMONIC_OR_PK: "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
```
wsdt, line 113: Should we have the deployer pk and other "public secrets" maybe in the .env file, to have a single source of truth?

ops/docker-compose-side.yml
```diff
@@ -69,3 +69,33 @@ services:
...
+      NETWORK: "http://l2geth:8545"
+      ENTRYPOINT: "0x"
+      BENEFICIARY: "0x"
+      HELPER: "0x"
+      ADDRESS_MANAGER_ADDRESS: "0x5FbDB2315678afecb367f032d93F642f64180aa3"
```
wsdt, line 118: same here - even if we are using create2 could definitely ease changes at some point. 
Big fan of having one configuration for the whole tech stack, makes it easy to use before people need to go into compose files and adapt things in case something changes at some point.

ops/docker-compose-side.yml
InoMurko: I don't think so - bad for services isolation. If you compromise one, you get access to everything ;). But anyway out of scope for this PR.

wsdt (approved):
Looks good to me, left a few comments.

ops/docker-compose-side.yml
wsdt: Isn't that the deployer key that we use for everything in the docker-compose file? in that case this would already be what we have now

InoMurko: Well you edited your comment in the mean time which changed the entire meaning haha. I thought you meant production services and AWS.

What you're proposing makes sense but only if each service's ENV files would have a prefix. Like `BUNDLER_PRIVATE_KEY`
`BUNDLER_MIN_BALANCE` etc.
otherwise you can have conflicting variable names.
In this case, `MNEMONIC_OR_PK` will be replaced by an unsused PK because it needs to submit transactions. But tests are not there yet so this PK will get replaced in a later PR.

wsdt: that's the old comment: `Isn't that the deployer key that we use for everything in the docker-compose file?`

boyuan-chen: Yes! We use different keys for different things. Keys has not been reused in different services.

boyuan-chen (approved):
LGTM

Issue #484, opened 2022-12-01
[DTL] Uploading a file with the wrong content-type silently errors


rupertbg:
## Issue Type

```
[x] bug report
[ ] feature request
```

## Current Behavior
When uploading a file to the DTL, setting the wrong content-type for the file does not throw an error. The file appears to upload but is not available for clients of the DTL.

Example:
`curl -H 'Content-Type: application/octet-stream' -T boba-addr.json http://dtl:8081/boba-addr.json`

The file will not be available but the curl will respond 200

## Expected Behavior
The example curl command should return an error code >400

## Steps to Reproduce

1. Upload a file to the DTL but with the wrong content-type header in the http request
2. No error will be thrown
3. The file will not be available from the DTL

## Suggested Fix
Check the content-type header corresponds to the file being uploaded and error if not correct.

PR #538, opened 2023-01-24
Validation for uploaded DTL address lists

mmontour1306:
Validation for uploaded DTL address lists, checking that the content-type was specified as application/json and that the list is not empty. Fixes issue #484 

 Changes to be committed:
	modified:   packages/data-transport-layer/src/services/main/service.ts


InoMurko (approved):

souradeep-das (approved):

boyuan-chen:
Could you please fix the indent?

boyuan-chen (approved):

PR #549, opened 2023-01-31
[AA] fix: entrypoint deployment reliability

### LINKED ISSUE #571: [AA] Make Entrypoint deploy deterministically
The Entrypoint contract should be deterministically deployed (to at least have the provision of having the contract on the same address across all Boba instances in the future)
network: mainnet,goerli

To determine - the deterministic deployer factory contract, either:
- use gnosis safe singleton factory as the deployer
- create a new deterministic deployer factory

souradeep-das:
:clipboard: assoicated: https://github.com/bobanetwork/boba/issues/571 https://github.com/bobanetwork/boba/issues/576

## Overview

Changes account-abstraction contract deploy files to NOT use hardhat-deploy to deploy contracts.
hardhat-deploy sometimes fails silently, (because of nonce issues)

deploys contracts deterministically (fixes: https://github.com/bobanetwork/boba/issues/571)
changes aa-sdk to work with boba (fixes: 2,3 from https://github.com/bobanetwork/boba/issues/576)

## Changes

- changes deploy scripts to a contractFactory pattern and deploys it through a deterministic deployment factory
- expands sdk to allow specifying network and on the basis of which determinstic deployment factory is selected
- expands sdk to allow specifying ethers.Wallet (provider.signer.sendTransaction is not supported on Boba)

## Testing

cd packages/account-abstraction
yarn deploy
(also check ci logs)


wsdt (approved):

wsdt (approved):

InoMurko:
Question to re: deterministic deployments.
It seems odd to me that the upstream repo would basically prevent deterministic deployments to chains that have protection.
What can be done about it?

My concerns are that if the entrypoint contracts become something like safe-deployments from Gnosis (standardizes addresses) and we ship this to prod with undeterministic addresses, it would make us uncompatible. Thoughts?

souradeep-das:
ohh upstream does have deploy scripts that uses the sdk - which has deterministic deploy through the [arachnid](https://github.com/Arachnid/deterministic-deployment-proxy) proxy(which we cant support - so entrypoint address might differ between mainnet and boba). We could still do a deterministic deploy using another factory contract - one that we deploy or we could use the factory we used for gnosis safes too. If we use the latter - we could have same addresses for all entrypoints across boba instances and in addition also other chains that choose to use the gnosis factory. 
since the addresses ultimately depend on the factory - chains with replay protection could end up having different addresses depending on the factory they use, but we can ensure that all our instances have the same address, until there is a standardized factory for these chains (like gnosis)
We could also have multiple entrypoints at the same time, but its easier for developers to follow if we can mark one as preferred and stick to that.

So i removed the deterministic part from the scripts with this change, but adding that part back in (#571 ) 
what deterministic deployer factory do we use? the [gnosis factory](https://github.com/safe-global/safe-singleton-factory)?

InoMurko (approved):

Issue #501, opened 2022-12-22
[ AA ]  examples as integration tests: user has no BOBA or ETH and dapp wants to sponsor ALL his transactions for their dapp


InoMurko:
## Issue Type

```
[ ] bug report
[x ] feature request
```

## Current Behavior
User has no $Boba or $Eeth and Dapp  developer wants to sponsor ALL users transactions only for their Dapp

PR #562, opened 2023-02-03
[AA] tests: add integrations tests for AA bundler, wallets, paymaster

### LINKED ISSUE #502: [ AA ] examples as integration tests: Firefly dapp wants users to use $FLY token to pay for transaction fees - L2 accepts BOBA
## Issue Type

```
[ ] bug report
[ x] feature request
```

## Current Behavior

Through account abstraction, a user that has $FLY ERC20 can use Boba L2. Dapp developer carries currency risk. We accept our native token (BOBA erc20 or Eth or Boba native or ERC20 destination chain (like glmr))

souradeep-das:
:clipboard: closes https://github.com/bobanetwork/boba/issues/502 #501 https://github.com/bobanetwork/boba/issues/576

## Overview

Adds integration tests for the Boba AA stack. Tests verify the following-
- aa-sdk is functional'
- bundler is able to submit operations and the operations are able to trigger changes
- simple wallet operations work through userOps
- paymasters work as intended
- boba token can be used as a fee token with a paymaster (BobaDepositPaymaster)
- any alt erc20 token can be used as a fee token with a paymaster (ManualDepositPaymaster)

## Changes

- Adds integration tests for the above mentioned cases
- fixes sdk issue 1,3 and 4 (https://github.com/bobanetwork/boba/issues/576)
- fixes bundler operation submission
- removes bundler side pre-simulation because it requires debug_traceCall, which is not supported on Boba

## Testing

cd integration-tests/
yarn test:integration


integration-tests/test/boba_aa_fee_alt_token.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+    console.log(L2ERC20Token.address)
+    await L2ERC20Token.connect(env.l2Wallet_2).mint(env.l2Wallet.address, utils.parseEther('500'))
+
+    bundlerProvider = new HttpRpcClient(
+      'http://localhost:3000/rpc',
+      entryPointAddress,
+      await env.l2Wallet.provider.getNetwork().then((net) => net.chainId)
+    )
+  })
+  it('{tag:other} should be able to send a userOperation to a wallet through the bundler', async () => {
```
InoMurko: tags can be dropped these days! fyi

integration-tests/test/boba_aa_fee_alt_token.spec.ts
```diff
@@ -0,0 +1,218 @@
+import chai from 'chai'
```
InoMurko: on the offchance that someone - a developer - looks at these integration tests - could you describe their intend a bit better? maybe from setup to what they actually plan to cover

integration-tests/test/boba_aa_fee_alt_token.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+
+      const OracleContract__factory = new ContractFactory(
+        MockFeedRegistryJson.abi,
+        MockFeedRegistryJson.bytecode,
+        env.l2Wallet
+      )
+
+      ManualDepositPaymaster = await ManualDepositPaymaster__factory.deploy(
+        entryPointAddress,
+      )
```
InoMurko: my question is - is this the right separation between `setup` and `test`?
and this applies for the entire suite of AA tests

integration-tests/test/boba_aa_fee_alt_token.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+    const preApproveDepositAmount = (await ManualDepositPaymaster.depositInfo(L2ERC20Token.address, account.address)).amount
+    const preApproveEtherBalance = await env.l2Provider.getBalance(account.address)
+
+    try {
+        const requestId = await bundlerProvider.sendUserOpToBundler(approveOp)
+        const txid = await walletAPI.getUserOpReceipt(requestId)
+        console.log('reqId', requestId, 'txid=', txid)
+    } catch (e) {
+        console.log(e)
+        throw new Error('Submission to Bundler Failed')
```
InoMurko: maybe throw e too?

integration-tests/test/boba_aa_fee_alt_token.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    const postApproveTokenBalance = await L2ERC20Token.balanceOf(account.address)
+    const postApproveDepositAmount = (await ManualDepositPaymaster.depositInfo(L2ERC20Token.address, account.address)).amount
+    const postApproveEtherBalance = await env.l2Provider.getBalance(account.address)
+    const op = await walletAPI.createUnsignedUserOp({
+      target: recipient.address,
+      data: recipient.interface.encodeFunctionData('something', ['hello']),
+      maxFeePerGas: '0x59682F00',
+      maxPriorityFeePerGas: '0x59682F00',
```
InoMurko: where do these constants come from?

integration-tests/test/boba_aa_sponsoring_fee.spec.ts
```diff
@@ -69,78 +69,90 @@ describe('Sponsoring Tx\n', async () => {
...
-      provider: env.l2Provider,
-      entryPointAddress,
-      owner: env.l2Wallet_3,
-      walletAddress: account.address,
+  describe('A user has no fee token, but pays for a transaction through a willing paymaster\n', async () => {
+    let walletAPI: SimpleWalletAPI
+    let signedOp
+    let account
+
+    it('should be able to add paymaster stake ', async () => {
```
souradeep-das: is splitting this into multiple assertions any better? or should i keep it under one block like `boba_aa_fee_alt_token.spec.ts`

integration-tests/test/boba_aa_fee_alt_token.spec.ts
souradeep-das: changed to make the sdk fetch gasPrice and use those values

integration-tests/test/boba_aa_fee_alt_token.spec.ts
souradeep-das: thanks, tried to refactor it. Related to this, wdyt about https://github.com/bobanetwork/boba/pull/562#discussion_r1114550985?

integration-tests/test/boba_aa_fee_alt_token.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+      data: recipient.interface.encodeFunctionData('something', ['hello']),
+    })
+
+
+    op.paymasterAndData = hexConcat([ManualDepositPaymaster.address, hexZeroPad(L2ERC20Token.address, 20)])
+    op.preVerificationGas = await walletAPI.getPreVerificationGas(op)
+
+    const signedOp = await walletAPI.signUserOp(op)
+
+    try {
```
boyuan-chen: Do we want to skip the test if anything goes wrong?

souradeep-das: the `sendUserOpToBundler()` operation seems to require a try catch, if it fails something is wrong - so lets make throw an error!

integration-tests/hardhat.config.ts
```diff
@@ -50,6 +50,17 @@ const config: HardhatUserConfig = {
           },
         },
       },
+      {
+        version: '0.8.12', // Required for WETH9
```
InoMurko, line 54: https://github.com/bobanetwork/boba/issues/668

packages/boba/bundler_sdk/src/calcPreVerificationGas.ts
```diff
@@ -67,6 +67,9 @@ export function calcPreVerificationGas (userOp: Partial<NotPromise<UserOperation
     ...userOp
   } as any
 
+  if (p.signature === '') {
+    p.signature = hexlify(Buffer.alloc(ov.sigSize, 1))
```
InoMurko, line 71: why is the dummy signature twice?
line 66 has dummy signature set - does that mean that the merge with userOp would overwrite it?

InoMurko: it makes sense then to remove the line on 66

packages/boba/bundler_sdk/src/Provider.ts
```diff
@@ -25,21 +26,30 @@ const debug = Debug('aa.wrapProvider')
 export async function wrapProvider(
   originalProvider: JsonRpcProvider,
   config: ClientConfig,
-  originalSigner: Signer = originalProvider.getSigner()
+  originalSigner: Signer = originalProvider.getSigner(),
```
InoMurko, line 31: do you want to update the `param` docs above the function definition?

InoMurko: why does it make sense to defer ownership to a wallet?

packages/boba/bundler/src/UserOpMethodHandler.ts
```diff
@@ -238,7 +238,7 @@ export class UserOpMethodHandler {
       )}`
     )
 
-    await this.simulateUserOp(userOp1, entryPointInput)
+    // await this.simulateUserOp(userOp1, entryPointInput)
```
InoMurko, line 242:
```suggestion
    //this is doing the geth debug_traceCall that the current ORU doesn't support
    // await this.simulateUserOp(userOp1, entryPointInput)
```

integration-tests/test/eth-l2/boba_aa_sponsoring_fee.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+      signedOp = await accountAPI.signUserOp(op)
+
+      expect(await signedOp.sender).to.be.eq(account.address)
+    })
+    it('should be able to submit the userOp to the bundler and trigger tx', async () => {
+      const preUserBalance = await env.l2Provider.getBalance(env.l2Wallet_4.address)
+      const prePaymasterDeposit = await VerifyingPaymaster.getDeposit()
+
+      try {
+        const requestId = await bundlerProvider.sendUserOpToBundler(signedOp)
```
InoMurko: I don't like the (hidden) dependency between tests on `signedOp`!  it seems  the previous test should be a setup(before_all) and the expect statement in a separate test?

integration-tests/test/eth-l2/boba_aa_sponsoring_fee.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+    let account
+
+    it('should be able to add paymaster stake ', async () => {
+      await VerifyingPaymaster.addStake(1, { value: utils.parseEther('2') })
+      await EntryPoint.depositTo(VerifyingPaymaster.address, {
+        value: utils.parseEther('1')
+      })
+
+      expect(await VerifyingPaymaster.getDeposit()).to.be.eq(utils.parseEther('1'))
+    })
```
InoMurko: why does this test belong in the group where the user doesn't have a fee token?

integration-tests/test/eth-l2/boba_aa_fee_boba.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+      EntryPointJson.abi,
+      env.l2Wallet
+    )
+
+    await BobaDepositPaymaster.addStake(1, { value: utils.parseEther('2') })
+    await EntryPoint.depositTo(BobaDepositPaymaster.address, {
+      value: utils.parseEther('1')
+    })
+  })
+  it('A user without ETH pays for a tx using $BOBA through a paymaster', async () => {
```
InoMurko: just to make things clear - ofc this does not mean it's using BOBA as the secondary fee token on the network

InoMurko: the paymaster could be accepting any token erc20 and pay for the userOp with it's own Eth as a wallet enabler

integration-tests/test/eth-l2/boba_aa_fee_boba.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+    const preApproveTokenBalance = await L2BOBAToken.balanceOf(account.address)
+    const preApproveDepositAmount = (await BobaDepositPaymaster.depositInfo(L2BOBAToken.address, account.address)).amount
+    const preApproveEtherBalance = await env.l2Provider.getBalance(account.address)
+
+    try {
+        const requestId = await bundlerProvider.sendUserOpToBundler(approveOp)
+        const txid = await accountAPI.getUserOpReceipt(requestId)
+        console.log('reqId', requestId, 'txid=', txid)
+    } catch (e) {
+        throw new Error('Submission to Bundler Failed: ' + e)
```
InoMurko: does this happen at all? in tests? would perhaps make more sense to just check on the txid with `expect` rather that a try catch imo

integration-tests/test/eth-l2/boba_aa_fee_boba.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+    const postApproveTokenBalance = await L2BOBAToken.balanceOf(account.address)
+    const postApproveDepositAmount = (await BobaDepositPaymaster.depositInfo(L2BOBAToken.address, account.address)).amount
+    const postApproveEtherBalance = await env.l2Provider.getBalance(account.address)
+    const op = await accountAPI.createUnsignedUserOp({
+      target: recipient.address,
+      data: recipient.interface.encodeFunctionData('something', ['hello']),
+    })
+
+
+    // TODO: check why paymasterAndData does not work when added to the walletAPI
```
InoMurko: you mentioned this is an optimisation, what does it give?

integration-tests/test/eth-l2/boba_aa_fee_boba.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+      expect(preApproveEtherBalance).to.gt(postApproveEtherBalance)
+      // users deposit amount on paymaster remains constant and is unused throughout
+      expect(preApproveDepositAmount).to.eq(postApproveDepositAmount)
+      expect(postApproveDepositAmount).to.eq(postCallDepositAmount)
+      // no ether is used when calling the recipient with the help of the paymaster, users boba token is used to pay
+      expect(postApproveEtherBalance).to.eq(postCallEtherBalance)
+      expect(postApproveTokenBalance).to.gt(postCallTokenBalance)
+      expect(BigNumber.from(postCallTokenBalance).add(logEP.args.actualGasCost)).to.closeTo(BigNumber.from(postApproveTokenBalance), utils.parseEther('0.0001'))
+    } catch (e) {
+      throw new Error('Submission to Bundler Failed: ' + e)
```
InoMurko: this is a huge try catch - I would rather see if we can avoid this

InoMurko (changes requested):
I would just suggest polishing a few things - like the big try/catch statements that might catch things out of their scope and then we're good!

boyuan-chen:
Can we add tests to the alt-l2 test folder?

integration-tests/test/boba_aa_sponsoring_fee.spec.ts
InoMurko: imo tests shouldn't be coupled - there shouldn't be a dependency between them. what do you think?

packages/boba/bundler_sdk/src/Provider.ts
souradeep-das: updated the param docs, thanks!

boba does not allow eth_sendTransaction from a remoteSigner, so we override using a wallet ([sdk issue-3](https://github.com/bobanetwork/boba/issues/576))

packages/boba/bundler/src/UserOpMethodHandler.ts
souradeep-das: added

integration-tests/test/eth-l2/boba_aa_sponsoring_fee.spec.ts
souradeep-das: rephrased the it statement
This step is where the paymaster operator funds the paymaster

integration-tests/test/eth-l2/boba_aa_fee_boba.spec.ts
souradeep-das: rephrased and added a comment

integration-tests/test/eth-l2/boba_aa_fee_boba.spec.ts
souradeep-das: it would allow to specify the paymasterAPI param in the walletAPI object from the sdk once and use it throughout. Right now we are self adding the 'paymaster' field to the constructed userOp, which can be prevented. makes it easier!

integration-tests/test/boba_aa_sponsoring_fee.spec.ts
souradeep-das: makes sense. 
The setup/it blocks looked huge and with a lot of steps, so tried to split them into multiple intermediary steps, but which would have these dependencies
Although i have put them under a separate new describe block, that contains all the dependent steps.

Do you think i should back to one it block? trading a bit of readability
associated: https://github.com/bobanetwork/boba/pull/562#discussion_r1143890610

InoMurko: I don't see a problem with a huge setup - it is what it is. 
Perhaps it makes sense to split them into separate files (even if the setup repeats). 

As long as tests are not coupled.

InoMurko:
> Can we add tests to the alt-l2 test folder?

that makes sense yeah!

InoMurko:
> > Can we add tests to the alt-l2 test folder?
> 
> that makes sense yeah!

you could do it in a follow up PR too though.

souradeep-das:
> Can we add tests to the alt-l2 test folder?

oh yes, the GPOPaymaster is specifically for alt-l2, will add that + aa_wallet test

integration-tests/test/boba_aa_sponsoring_fee.spec.ts
souradeep-das: okayy

integration-tests/test/eth-l2/boba_aa_fee_boba.spec.ts
souradeep-das: dont know why i need this before for the sendUserOptoBundler operation, but it works without now - removed

integration-tests/test/eth-l2/boba_aa_fee_boba.spec.ts
souradeep-das: rmd

integration-tests/test/boba_aa_sponsoring_fee.spec.ts
souradeep-das: how about now ?

packages/boba/bundler_sdk/src/calcPreVerificationGas.ts
souradeep-das: L66 is for when userOp.signature doesn't exist, and L71 for when its empty. 
I can combine them together, but is keeping this as an addition better for pulling changes in the future?

souradeep-das:
moving the alt-l1 tests to a new PR, because it involves tests for a new contract
tracked here: https://github.com/bobanetwork/boba/issues/670

InoMurko (approved):

boyuan-chen (approved):

PR #600, opened 2023-02-23
Add rate limit in proxy for websocket

boyuan-chen:

## Overview

- Websocket uses the same IP rate limit system as the HTTP port
- Websocket has the same sender rate limit system as the HTTP port

## Changes

- Add rate limit checks in the backend service
- Add sender rate limit checks in the backend service
- Add maxResponseSize and maxBodySize

## Testing

Add unit tests


InoMurko (approved):

go/proxyd/integration_tests/testdata/ws.toml
```diff
@@ -4,7 +4,8 @@ ws_backend_group = "main"
 
 ws_method_whitelist = [
   "eth_subscribe",
-  "eth_accounts"
+  "eth_accounts",
+    "eth_sendRawTransaction"
```
InoMurko, line 8: indent

InoMurko:
```suggestion
  "eth_accounts",
  "eth_sendRawTransaction"
```

InoMurko (commented):
@Boyuan-Chen rate limiting that was implemented on multiple levels (per account, ip, rpc, websockets) - does it permit to be forwarded to a configured backend (alchemy, tenderly) without rate limit?

these backends will most probably have their own rate limiting enabled that would largely be completely different to ours.

boyuan-chen:
> @Boyuan-Chen rate limiting that was implemented on multiple levels (per account, ip, rpc, websockets) - does it permit to be forwarded to a configured backend (alchemy, tenderly) without rate limit?
> 
> these backends will most probably have their own rate limiting enabled that would largely be completely different to ours.

No, but I don't think we should allow unlimited requests to a configured backend (alchemy, tenderly). I assume that the key we get from alchemy and tenderly has a very high rate limit, so we shouldn't let others use it as free or do the free DOS to their system. If they want to have a higher rate limit, they should buy keys directly from alchemy or tenderly.

There are multiple ways for us to be excluded with a specific header that allows us to have the unlimited rate.

InoMurko:
>  
> 
> 
> 
> > @Boyuan-Chen rate limiting that was implemented on multiple levels (per account, ip, rpc, websockets) - does it permit to be forwarded to a configured backend (alchemy, tenderly) without rate limit?
> 
> > 
> 
> > these backends will most probably have their own rate limiting enabled that would largely be completely different to ours.
> 
> 
> 
> No, but I don't think we should allow unlimited requests to a configured backend (alchemy, tenderly). I assume that the key we get from alchemy and tenderly has a very high rate limit, so we shouldn't let others use it as free or do the free DOS to their system. If they want to have a higher rate limit, they should buy keys directly from alchemy or tenderly.
> 
> 
> 
> There are multiple ways for us to be excluded with a specific header that allows us to have the unlimited rate.

We'll see what kind of agreement we have then

boyuan-chen:
> 



> > > @Boyuan-Chen rate limiting that was implemented on multiple levels (per account, ip, rpc, websockets) - does it permit to be forwarded to a configured backend (alchemy, tenderly) without rate limit?
> > 
> > 
> > > 
> > 
> > 
> > > these backends will most probably have their own rate limiting enabled that would largely be completely different to ours.
> > 
> > 
> > No, but I don't think we should allow unlimited requests to a configured backend (alchemy, tenderly). I assume that the key we get from alchemy and tenderly has a very high rate limit, so we shouldn't let others use it as free or do the free DOS to their system. If they want to have a higher rate limit, they should buy keys directly from alchemy or tenderly.
> > There are multiple ways for us to be excluded with a specific header that allows us to have the unlimited rate.
> 
> We'll see what kind of agreement we have then

Ok! In the proxy.toml, there is a private key setting for the backend service. Different service providers might use different ways to authenticate, but the setting is there.

go/proxyd/integration_tests/ws_test.go
```diff
@@ -299,3 +256,143 @@ func TestWSClientMaxConns(t *testing.T) {
...
+		client, err := NewProxydWSClient("ws://127.0.0.1:8546", h, func(msgType int, data []byte) {
+			clientHdlr.MsgCB(msgType, data)
+		}, nil)
+		defer client.HardClose()
+		require.NoError(t, err)
+		res := spamWSReqs(t, clientHdlr, backendHdlr, client, sampleRequest, []byte(""), 4)
+		require.Equal(t, 0, res[invalidRateLimitResponse])
+	})
+
+	t.Run("exempt over limit", func(t *testing.T) {
```
mmontour1306: It looks like this "exempt over limit" test is the same code as "exempt user agent over limit" above. The corresponding section of rate_limit_test.go inserts an Origin header here rather than User-Agent.

go/proxyd/integration_tests/testdata/frontend_rate_limit.toml
```diff
@@ -26,4 +27,9 @@ error_message = "over rate limit with special message"
...
 [rate_limit.method_overrides.eth_foobar]
 limit = 1
-interval = "1s"
\ No newline at end of file
+interval = "1s"
+
+[rate_limit.method_overrides.eth_baz]
+limit = 1
+interval = "1s"
+global = true
```
mmontour1306, line 35: I see a few "no newline at end of file" warnings such as here. I don't know if there's an official policy on that but I try to avoid it.

mmontour1306 (approved):

go/proxyd/integration_tests/ws_test.go
boyuan-chen: Fixed. I can only set the origin header to the URL. Otherwise, the websocket package rejects the connection.
```
// In gorilla/websocket, the Origin header must be the same as the URL.
// Otherwise, it will be rejected
h.Set("Origin", "wss://127.0.0.1:8546")
```

go/proxyd/integration_tests/testdata/frontend_rate_limit.toml
boyuan-chen: It seems like Optimism adds a new line in some files, but some files are not added a new line. I will create a PR and see what they want us to do.

InoMurko:
is this ready to be merged?

PR #613, opened 2023-03-01
[AA] unify aa-contracts package

souradeep-das:
:clipboard:

## Overview

exports all AA-contract artifacts and factories from one package

## Changes

- pack and build from @boba/accountabstraction instead of packing  contracts/*
- exports all contract factories from the package, so files that used @account-abstraction/contracts can simply use @boba/accountabstraction instead (only artifacts structure is distinct to the other package - it is the default hardhat compile structure)

## Testing

check ci logs for aa_deployment


InoMurko (approved):

packages/boba/account-abstraction/package.json
```diff
@@ -24,6 +27,14 @@
     "runop-goerli": "AA_URL=https://account-abstraction-goerli.nethermind.io yarn runop --network goerli",
     "runop3": "hardhat run src/runop3.ts "
   },
+  "keywords": [
+    "solidity",
+    "ethereum",
+    "smart",
+    "contracts",
+    "accipt-abstraction",
+    "eip-4337"
```
InoMurko:
```suggestion
    "account-abstraction",
    "eip-4337"
```

PR #531, opened 2023-01-21
sideload bundler and add address manager

InoMurko:
[:clipboard: Add associated issues, tickets, docs URL here.](https://github.com/bobanetwork/boba/issues/504)

## Overview
Shows how to build Bundler service and loads it in the docker compose side. Integrates our Address Manager to retrieve deployed contracts done in a separate pr from @souradeep-das  

## Changes

Describe your changes and implementation choices. More details make PRs easier to review.

- Alpine image with two layers
- Import address manager from contracts and use it to retrieve two registered contracts if address manager is present

## Testing

Will add more tests once things are tied together with @souradeep-das

PR #659, opened 2023-03-14
Merge alt l1 to develop

boyuan-chen:
## Overview

Merge alt-l1 branch to develop branch

## Changes

- Add a flag in l2geth to enable the node for alt l2s.
- Add subgraph supports in SDK and DTL for alt l1s
- Add core contracts for Alt L1s to the `packages/contracts/`
- Modify core deployer to support alt l1s
- Add boba contracts for Alt L2s to the `packages/boba/contracts
- Modify boba deployer to support alt L2s
- Add integration tests for alt l1s to `integration-tests/test/alt-l2` folder
- Add MaxGasLimit in batch submitter for moonbeam
- Add integration tests for all networks
- Fix minor issues

## Testing

/


packages/contracts/src/index.ts
```diff
@@ -1,4 +1,3 @@
 export * from './contract-defs'
 export * from './predeploys'
 export * from './connect-contracts'
-export * from './local-network-config'
```
boyuan-chen: This fixes the problem of breaking the gateway.

sk-enya: aah awesome!! good finding....

boyuan-chen: Details: exporting this 
https://github.com/bobanetwork/boba/blob/a35f9e1c63778ebb10ad7377871ca0cbd53a8f11/packages/contracts/src/local-network-config.ts#L1-L15
to the root directory breaks the gateway. We should not let the gateway access the `hardhat` package.

sk-enya: Hmm, not sure how to do that but what I know as per my understanding gateway mostly import the `node_modules` from root as far as it's common.

boyuan-chen: Yup! I don't know why the gateway breaks when it can access something inside `hardhat/internal/core`.

.circleci/config.yml
```diff
@@ -79,109 +183,63 @@ jobs:
...
-      - run:
-          name: Bring up the stack
-          command: |
-            docker images
-            BUILD=2 DAEMON=1 ./up_local.sh
-          working_directory: ops
-      - run:
-          name: Bring up services
-          command: |
-            docker-compose -f docker-compose.yml -f docker-compose-side.yml up -d bobalink aa_deployer bundler
```
souradeep-das, line 151: a part of building and spinning up the bundler got removed, can we add this back? 😄

boyuan-chen: The bundler test is already added.
https://github.com/bobanetwork/boba/blob/5afcc237f5cf38f5b8f3b78450de54f60c8153e0/.circleci/config.yml#L160

boyuan-chen: Did I miss anything?

souradeep-das: we are now building the bundler (as a side service) and spinning it up with the stack - to enable the integration tests to work 😄, that part seems to have been removed
so just adding the bundler to docker-compose build and up 
(ref - the pr that added this https://github.com/bobanetwork/boba/pull/531/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47)

InoMurko: let me take a stab at this

InoMurko (approved):

PR #692, opened 2023-04-05
Performance updates

alvaro-ricotta:
1. We have converted all regular lodash imports to deepimports and move to 'util/lodash' then exporting the methods directly from 1 file, which allows the bundlesize to significantly reduce its size.

2. I have made a list of functions available in util/dates to avoid re-importing libraries from various files. Now, the moment library is only imported from the file where we export all the necessary functions to handle dates. This results in a significant increase in bundle size as well as an improvement in execution and rendering times.

3. Moment.js has been replaced by Day.js, which is 30 times lighter and much faster in terms of response time, and in certain cases, up to 50% more efficient.

**Explanation:** 
The technical explanation behind this is how module systems work in JavaScript and how dependencies are handled in an application.

When modules are imported in JavaScript, the load and compilation time of the application can be affected if many dependencies are being repetitively imported in different files. This is because every time a module is imported, a search and load is performed in memory for that particular module. If many files import the same module, then that module is loaded and compiled multiple times, which can negatively affect performance.

By moving imports to specific files and exporting them from there, a module is created that encapsulates all those shared and reusable functionalities throughout the application. This way, the number of times those modules need to be searched for and loaded in memory is reduced, as they are only imported from a single place.


sk-enya (approved):
Looks Good!

boyuan-chen (approved):

PR #698, opened 2023-04-10
Inomurko/bump bundler 

InoMurko:
This PR brings in 0.5.0 Bundler changes from upstream.
It also introduces a wrapper contract `EntryPointWrapper` for the purpose of EntryPoint interaction through custom errors that our l2geth currently does not support.

## Overview

Re-enabled and fixed tests:
boba_aa_wallet.spec
boba_aa_sponsoring_fee.spec
boba_aa_fee_boba.spec
boba_aa_fee_alt_token.spec
## Changes

- Renaming everything from Wallet to Account
- I think GethTracer could be deleted, but post Bedrock it might be useful (or even with Tenderly now)
- Brings in mempool support
- 
## Testing

Unit tests and integration tests are in the PR


integration-tests/test/eth-l2/boba_aa_wallet.spec.ts
```diff
@@ -1,209 +1,222 @@
...
+      recipient.address,
+      'Sender'
+    )
+    const log = recipient.interface.parseLog(receipt.logs[returnedlogIndex])
+    // tx.origin is the bundler
+    expect(log.args.txOrigin).to.eq(env.l2Wallet.address)
+    // msg.sender is the 4337 wallet
+    expect(log.args.msgSender).to.eq(account.address)
+    // message is received and emitted
+    expect(log.args.message).to.eq('hello')
```
InoMurko: @souradeep-das this doesn't make sense anymore, right?
```
expect(await op.sender).to.be.eq(senderCreator.address)
```
if op.sender is created on the fly, I can't check it.

What about ?
```
expect(log.args.msgSender).to.eq(account.address)
```

souradeep-das: I think for this specific test case we were deploying the account ourselves?
so we could deploy the account and then pass it on to the sdk such that it is not deployed by the signedUserOp with something like
```
const deployedAccount = await account.createAccount(env.l2Wallet.address, 0)

const accountAPI = new SimpleAccountAPI({ 
   provider: env.l2Provider,
   entryPointAddress: entryPointAddress,
   senderCreatorAddress: senderCreator.address,
   owner: env.l2Wallet,
   accountAddress: deployedAccount.address
})
```
<br><br>

If however, we were not deploying the op.sender (which was the case with a test case) and its being created on the fly with the userOp -
we could get the address through `accountAPI.getAccountAddress()` since it is to be deterministically deployed through the factory
then op.sender and log.args.msgSender can be compared to this returned account?

packages/boba/bundler/contracts/tests/TestStorageAccount.sol
```diff
@@ -0,0 +1,73 @@
...
+import "@boba/accountabstraction/contracts/interfaces/IAccount.sol";
+import "@boba/accountabstraction/contracts/interfaces/IPaymaster.sol";
+import "@boba/accountabstraction/contracts/interfaces/IEntryPoint.sol";
+import "./TestRuleAccount.sol";
+import "./TestCoin.sol";
+
+/**
+ * an account with "rules" to trigger different opcode validation rules
+ */
+contract TestStorageAccount is TestRuleAccount {
```
wsdt, line 13: are these actual test (helpers/mocks)? 
Out of scope, but have we ever thought about adding Foundry tests for cases like this? 
I find Foundry speeds up development efforts

InoMurko: These are tests, but from upstream, like 95% of this PR

wsdt: ok thx!

packages/boba/account-abstraction/contracts/bundler/EntryPointWrapper.sol
```diff
@@ -0,0 +1,208 @@
...
+    AggregatorStakeInfo private emptyAggregatorInfo = AggregatorStakeInfo(address(0), emptyStakeInfo);
+    ReturnInfo private emptyReturnInfo = ReturnInfo(0, 0, false, 0, 0, new bytes(0));
+    Response private emptyResponse = Response("", emptyReturnInfo, emptyStakeInfo, emptyStakeInfo, emptyStakeInfo, emptyAggregatorInfo);
+    FailedOpStatus private emptyFailedOp = FailedOpStatus(false, 0, "");
+
+    constructor(IEntryPoint _entryPoint) {
+        entryPoint = _entryPoint;
+    }
+
+    function simulateValidation(UserOperation calldata userOp) external returns (FailedOpStatus memory, Response memory) {
```
wsdt, line 105: one thing I noticed when using this why we don't have the same return types for this and the actual entryPoint simulateValidation.

packages/boba/bundler/src/BundlerServer.ts
```diff
@@ -37,18 +51,47 @@ export class BundlerServer {
...
+      preVerificationGas: 0,
+      verificationGasLimit: 100000,
+      callGasLimit: 0,
+      maxFeePerGas: 0,
+      maxPriorityFeePerGas: 0,
+      signature: '0x',
+    }
+    // await EntryPoint__factory.connect(this.config.entryPoint,this.provider).callStatic.addStake(0)
+
+    // const err = await EntryPoint__factory.connect(
```
wsdt: do we need this?

packages/boba/bundler/src/BundlerServer.ts
```diff
@@ -119,8 +166,52 @@ export class BundlerServer {
           params[1]
         )
         break
+      // custom errors post bedrock!
+      // case 'eth_estimateUserOperationGas':
+      //   result = await this.methodHandler.estimateUserOperationGas(
```
wsdt: shouldn't we instead return an error that informs the user that the method exists but is not currently supported?

wsdt: otherwise it's just the generic not supported/not exists error

packages/boba/bundler/src/GethTracer.ts
```diff
@@ -58,20 +46,26 @@ export async function debug_traceTransaction (
...
- *
  * @param func
  */
 export function getTracerBodyString (func: LogTracerFunc): string {
   const tracerFunc = func.toString()
   // function must return a plain object:
   //  function xyz() { return {...}; }
-  const regexp =
-    /function \w+\s*\(\s*\)\s*{\s*return\s*(\{[\s\S]+\});?\s*\}\s*$/ // (\{[\s\S]+\}); \} $/
+  const regexp = /function\w*\s*\(\s*\)\s*{\s*return\s*(\{[\s\S]+\});?\s*\}\s*$/ // (\{[\s\S]+\}); \} $/
```
wsdt, line 55: not sure where else we have regex, but I think having these things in a global constants file for consistent behavior could be a good idea for predictability

wsdt (approved):
left some comments

packages/boba/account-abstraction/contracts/bundler/EntryPointWrapper.sol
InoMurko: can you elaborate?

the return of entrypoint does not matter much, because all code paths use custom `revert` (`ValidationResultWithAggregation`, `ValidationResult`, `FailedOp`)

packages/boba/bundler/src/GethTracer.ts
InoMurko: this isn't used because we can't trace on l2geth

packages/boba/account-abstraction/contracts/bundler/EntryPointWrapper.sol
wsdt:
```ts
let errorResult = await this.entryPointWrapper.callStatic
      .simulateValidation(userOp)
      .catch((e) => e)

    // Parse result from EntryPointWrapper
    errorResult = {
      errorName: errorResult[1][0],
      errorArgs: errorResult[1],
    }
```

Had to map the callstatic result to be consistent with the regular entryPoint. 
So not specifically the returnType

packages/boba/bundler/src/GethTracer.ts
InoMurko: also, there's a tradeoff between how upstream code looks like and what we want to do with it. since the plan is to pull changes from upstream often, I change only things that I have to.

packages/boba/bundler/src/BundlerServer.ts
InoMurko: https://github.com/eth-infinitism/bundler/blob/v0.5.0/packages/bundler/src/BundlerServer.ts#L74-L79

we might want to support this via the Wrapper contract, currently we cannot run this!

packages/boba/bundler/src/GethTracer.ts
wsdt: ah yeah I'm always just reviewing under the assumption that this is our code. :-) 
Sure with you, keep changes at minimum

packages/boba/bundler/src/BundlerServer.ts
InoMurko: https://github.com/bobanetwork/boba/issues/759

packages/boba/bundler/src/BundlerServer.ts
souradeep-das: i think this is the same endpoint you are working on? so we can add this back? "estimateUserOperationGas"

.circleci/config.yml
```diff
@@ -79,7 +79,7 @@ commands:
       - run:
           name: Bring up services
           command: |
-            docker-compose -f <<parameters.docker_compose_file>> -f docker-compose-side.yml up -d bobalink aa_deployer
+            docker-compose -f <<parameters.docker_compose_file>> -f docker-compose-side.yml up -d bobalink aa_deployer bundler
           working_directory: ops
       - run:
           name: Start background logging
```
souradeep-das, line 85: L257 - lets uncomment test_flow_bundler_and_depcheck?

packages/boba/bundler/src/BundlerServer.ts
InoMurko: yeah, that can be fixed later on. for now I added a new error type and throw an error for not supported

packages/boba/bundler/src/runner/runop.ts
```diff
@@ -2,85 +2,91 @@
...
 import { erc4337RuntimeVersion } from '@boba/bundler_utils'
 import fs from 'fs'
-import { HttpRpcClient } from '@boba/bundler_sdk/dist/src/HttpRpcClient'
-import { SimpleAccountAPI } from '@boba/bundler_sdk'
-import { DeterministicDeployer } from '@boba/bundler_sdk/dist/src/DeterministicDeployer'
+import {
+  DeterministicDeployer,
+  HttpRpcClient,
+  SimpleAccountAPI,
+} from '../../../bundler_sdk/dist'
```
souradeep-das: external dep (although isok, i dont think runop is used extensively)

packages/boba/bundler_sdk/test/0-deterministicDeployer.test.ts
```diff
@@ -32,11 +29,11 @@ describe('#deterministicDeployer', () => {
     await ethers.provider.getSigner().sendTransaction({ to: testWallet.address, value: ethers.utils.parseEther('0.01') })
 
     const ctr = hexValue(new SampleRecipient__factory(ethers.provider.getSigner()).getDeployTransaction().data!)
-    DeterministicDeployer.init(ethers.provider, testWallet)
+    console.log(DeterministicDeployer.init(ethers.provider, testWallet))
```
souradeep-das: maybe we want to remove log?
(here and L36)

integration-tests/test/eth-l2/boba_aa_fee_alt_token.spec.ts
```diff
@@ -1,220 +1,235 @@
...
+      // this approve operation needs gas (in eth) because this step does not involve a paymaster
+      await env.l2Wallet.sendTransaction({
+        value: utils.parseEther('2'),
+        to: account,
+      })
+
+      accountAPI = new SimpleAccountAPI({
+        provider: env.l2Provider,
+        entryPointAddress,
+        senderCreatorAddress: senderCreator.address,
```
souradeep-das: minor: since we are passing accountAddress, do we need to pass senderCreator additionally?
( if we dont need it we could remove it from L141)
(applies to the other tests files)

integration-tests/test/eth-l2/boba_aa_fee_alt_token.spec.ts
```diff
@@ -1,220 +1,235 @@
...
+    before('the paymaster operator sets up the paymaster by staking and adding deposits', async () => {
+      await ManualDepositPaymaster.addStake(1, { value: utils.parseEther('2') })
+      await EntryPoint.depositTo(ManualDepositPaymaster.address, {
+        value: utils.parseEther('1')
+      })
+    })
+
+    before('the user approves the paymaster to spend their $BOBA token', async () => {
+      // deploy a 4337 Wallet and send operation to this wallet
+      SimpleAccount__factory = new ContractFactory(
```
souradeep-das: super minor: should we call it 'SimpleAccountFactory__factory'?

(applies to the other tests files)

integration-tests/test/eth-l2/boba_aa_fee_alt_token.spec.ts
```diff
@@ -1,220 +1,235 @@
...
+        SampleRecipientJson.abi,
+        recipient.address,
+        'Sender'
+      )
+      const log = recipient.interface.parseLog(receipt.logs[returnedlogIndex])
+      // tx.origin is the bundler
+      expect(log.args.txOrigin).to.eq(env.l2Wallet.address)
+      // msg.sender is the 4337 wallet
+      console.log(log.args.msgSender)
+      console.log(account)
```
souradeep-das: unless this was added on purpose :D

integration-tests/test/eth-l2/boba_aa_wallet.spec.ts
```diff
@@ -1,209 +1,220 @@
...
+    const senderCreator = await SenderCreator__factory.deploy({ gasLimit: 9_500_000 })
+
+    console.log('Sender Creator Factory deployed to:', senderCreator.address)
+
+    const accountAPI = new SimpleAccountAPI({
+        provider: env.l2Provider,
+        entryPointAddress,
+        senderCreatorAddress: senderCreator.address,
+        owner: env.l2Wallet,
+        factoryAddress: accountFactory.address,
```
souradeep-das: lets remove factoryAddress and senderCreatorAddress here since we specify accountAddress?

integration-tests/test/eth-l2/boba_aa_wallet.spec.ts
```diff
@@ -1,209 +1,220 @@
...
+    const op = await accountAPI.createSignedUserOp({
+      target: recipient.address,
+      data: recipient.interface.encodeFunctionData('something', ['hello']),
+    })
+    await env.l2Wallet.sendTransaction({
+      value: utils.parseEther('2'),
+      to: await op.sender,
+    })
+
+    //what is this supposed to be?
```
souradeep-das: minor: a comment we can rem now 😄

integration-tests/test/eth-l2/boba_aa_wallet.spec.ts
```diff
@@ -1,209 +1,220 @@
...
+    await env.l2Wallet.sendTransaction({
+      value: utils.parseEther('2'),
+      to: accountAddress,
+    })
+
+    const op = await accountAPI.createSignedUserOp({
+      target: recipient.address,
+      data: recipient.interface.encodeFunctionData('something', ['hello']),
+    })
+    await env.l2Wallet.sendTransaction({
```
souradeep-das: minor: we are sending eth two times? on L181 too?

souradeep-das (approved):
this is super, lets go 🚀

.circleci/config.yml
InoMurko: this actually does not exist in upstream anymore! will remove completely

packages/boba/bundler/src/runner/runop.ts
InoMurko: resolved nevertheless!

PR #713, opened 2023-04-15
Fix: typos

omahs:
Fix: typos


InoMurko:
Thank you for your contribution! The documentation is in the branch `docs-in-monrepo`, if you could apply these changes there, that would be really helpful! Thanks

PR #722, opened 2023-04-19
simplify import from components + moving functions to ts

alvaro-ricotta:
An index.ts was created in the components folder, which should export all the components to simplify and reduce the imports required from the internal views.

import {Copy, Disconnect} from 'components'

In this way, we simplify imports in each view, reducing them only to the base file that will export all the general structures of the components. It is pending to apply the same logic for services, selectors, reducers, store, etc.

Several components were converted to ts, applying typing.


sk-enya (changes requested):
Failing with multiple error on local please refer screenshot below.
Screenshot 2023-04-20 at 3 26 28 PM
Screenshot 2023-04-20 at 3 27 04 PM

sk-enya:
@alvaro-ricotta please make sure to run `yarn && yarn clean && yarn build` from the root so we have updated yarn.lock corresponding to packages.

sk-enya:
@alvaro-ricotta if you want to skip executing ci for your commits you can pass any of the below string in commit message.

`[skip ci]`
`[ci skip]`
`[no ci]`
`[skip actions]`

sk-enya (approved):

InoMurko (approved):
```
+124 −477
```
:-)

PR #727, opened 2023-04-21
Audit updates qsp

souradeep-das:
updates to contracts (Quantstamp)


packages/boba/contracts/contracts/ERC1155Bridges/L1ERC1155Bridge.sol
```diff
@@ -4,7 +4,7 @@
 /**
  Note: This contract has not been audited, exercise caution when using this on mainnet
  */
-pragma solidity >0.7.5;
+pragma solidity 0.8.9;
```
InoMurko, line 7: hehe finally!

wsdt: also like the specific version definition

packages/boba/contracts/contracts/ERC1155Bridges/L2ERC1155BridgeAltL1.sol
```diff
@@ -75,6 +76,24 @@ contract L2ERC1155BridgeAltL1 is iL2ERC1155BridgeAltL1, CrossDomainEnabled, ERC1
     // billing contract address
     address payable public billingContractAddress;
 
+    event OwnershipTransferred(
+        address newOwner
```
wsdt, line 80: Any reason why we tend not to use OpenZeppelin and implement standard patterns such as Ownable ourselves? 
Or is this code also based on a "fork"?

wsdt (approved):
looks good in general, left some comments.

packages/boba/contracts/contracts/ERC1155Bridges/L2ERC1155BridgeAltL1.sol
souradeep-das: yeah that would have been better! But since we did not do in the beginning - i'm sticking with the structure because its under a proxy

wsdt: ok thx - makes sense!

packages/boba/contracts/contracts/ERC1155Bridges/L2ERC1155BridgeAltL1.sol
```diff
@@ -366,17 +400,13 @@ contract L2ERC1155BridgeAltL1 is iL2ERC1155BridgeAltL1, CrossDomainEnabled, ERC1
 
         PairTokenInfo storage pairToken = pairTokenInfo[_l2Contract];
         require(pairToken.l1Contract != address(0), "Can't Find L1 token Contract");
+        require(_from == msg.sender, "Sender does not have token priviledges");
 
         require(_amount > 0, "Amount should be greater than 0");
 
         if (pairToken.baseNetwork == Network.L1) {
-            address l1Contract = IL2StandardERC1155(_l2Contract).l1Contract();
-            require(pairToken.l1Contract == l1Contract, "L1 token Contract Address Error");
```
boyuan-chen, line 374: I think we want to cover this case by adding this check. What if developers accidentally update the l1Contract and we skip this check? The transaction will be reverted on l1.

souradeep-das: yeah - this is what i was thinking-
a) such a token contract that allowed l1Contract changing would be fraudulent and should not be added since it would make the previous tokens locked forever. 
b) And with this change i'm using `pairToken.l1Contract` instead of `l1Contract` when sending a message, such that the message is sent with the l1Contract registered on the bridge contract instead (such that registration takes precedence)

what do you think? lmk if we still want to add this back

boyuan-chen: oh, got it. That's great. We should add this extra note to our docs.

souradeep-das: perfect, so keeping this change then!
also added some notes on the user docs here https://github.com/bobanetwork/boba/pull/729

boyuan-chen (approved):

Issue #739, opened 2023-04-26
Publish Bundler SDK to NPM


wsdt:
https://github.com/bobanetwork/boba/pull/698

wsdt:
account creation in progress (Boba Foundation)

wsdt:
secret named `NPM_BOBA_FOUNDATION`

PR #749, opened 2023-04-27
Bundler SDK - NPM release pipeline

wsdt:
## Overview

Add CI/CD pipeline for `@bobanetwork/bundler_sdk`. 

resolves #[739](https://github.com/bobanetwork/boba/issues/739)

## Changes

- Rename bundler_sdk scope from `@boba` to `@bobanetwork` as the scope on npmjs is unfortunately already taken & refactored usages (but didn't do so for other packages yet - TBD)
- NPM package released once we create a new release automatically

## Testing

Test run of pipeline successful. Test package has been deleted from the registry.

### DIFF
--- a/.github/workflows/aa-bundler-sdk.yml
+++ b/.github/workflows/aa-bundler-sdk.yml
@@ -1,0 +1,23 @@
+name: Publish AA-Bundler to NPM
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+      - name: Install dependencies and build 🔧
+        run: yarn install --frozen-lockfile --legacy-peer-deps && yarn run build
+      - name: Publish package on NPM 📦
+        working-directory: ./packages/boba/bundler_sdk
+        run: npm publish --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_BOBA_FOUNDATION }}


souradeep-das (approved):
🚀

.github/workflows/aa-bundler-sdk.yml
```diff
@@ -0,0 +1,23 @@
...
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+      - name: Install dependencies and build 🔧
+        run: yarn install --frozen-lockfile --legacy-peer-deps && yarn run build
+      - name: Publish package on NPM 📦
+        working-directory: ./packages/boba/bundler_sdk
+        run: npm publish --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_BOBA_FOUNDATION }}
```
InoMurko, line 23: this probably takes `"version": "0.2.3",` from the package.json, right?
is there a way to prevent overwriting packages in NPM?

something like this: https://github.com/ethereum-optimism/optimism/pull/5413#issuecomment-1503778609

image

InoMurko (changes requested):

.github/workflows/aa-bundler-sdk.yml
wsdt: with the current configuration we only create a new package on new "Github release", not on new commit to a specific branch. So yes, right now we would need to update the version number of package.json. 

If you keep the same version number in the package.json then the release simply fails, and is not released. Basically npmjs.org prevents you from overwriting existing packages. 

@InoMurko

InoMurko: okay, makes sense. thats good to know.

should we address the second part of my comment separately - the automation (changeset PRs)?

wsdt: let's do it in a separate issue / PR because I guess we also might need to add a version check to avoid failing releases if possible (or auto-incrementing the version number etc).

wsdt: let's do it in a separate issue / PR because I guess we also might need to add a version check to avoid failing releases if possible (or auto-incrementing the version number etc).

wsdt: if we do auto-increment for the version, then we maybe could also consider releasing on commit (with changeset) as in your screenshot above.

InoMurko: yeah, one by one! open an issue pls!

wsdt: https://github.com/bobanetwork/boba/issues/780

InoMurko:
Please clean the commit history

InoMurko (approved):

PR #775, opened 2023-05-03
ValidationManager account for signature expiration

### LINKED ISSUE #753: [AA] ValidationManager needs to account for expiring UserOp
in `ValidationManager.ts`
```
// requireCond(
    //   res.returnInfo.validUntil == null ||
    //     res.returnInfo.validUntil + 30 < Date.now() / 1000,
    //   'expires too soon',
    //   ValidationErrors.ExpiresShortly
    // )
```

this is currently being ignored - but we shouldn't

wsdt:

## Overview

ValidationManager needs to account for expiring UserOp
Resolves #753

## Changes

- Reversing >< as probably semantic mistake. 
- Added `validAfter` requireCondition as it was missing altogether
- Added two integration tests to cover `validAfter` and `validUntil`
- Adding two unit tests to cover `validAfter` and `validUntil` (in progress)

## Testing

AA + bundler tests run through.


wsdt:
might add a dedicated test for expiration

InoMurko:
Okay - so this needs a bit more then uncommenting code. 

First, you can see there's no test coverage for this code path, not in unit tests and not in integration tests. Once that's added we'll have more understanding of the PR.

InoMurko:
The reason why this is more complex is the divergence of AA contracts and bundler. @souradeep-das has more to share here.

packages/boba/bundler/src/modules/ValidationManager.ts
```diff
@@ -162,17 +162,16 @@ export class ValidationManager {
...
-//       'expires too soon',
-//       ValidationErrors.ExpiresShortly
-//     )
-console.log(`res.aggregatorInfo ${res.aggregatorInfo}`)
-console.log(res.aggregatorInfo.addr)
-console.log(res.aggregatorInfo.stake)
-console.log(res.aggregatorInfo.unstakeDelaySec)
+    requireCond(
+      res.returnInfo.validUntil == null ||
+        res.returnInfo.validUntil + 30 > Date.now() / 1000,
```
InoMurko: Good catch

souradeep-das: great!

souradeep-das: I think this should be? -
```
res.returnInfo.validUntil > Date.now() / 1000 + 30
```

souradeep-das (commented):
great, I think the switch of the operator should make it work! (more precisely maybe the expression added as comment?)

this condition comes up in verifyingPaymaster/BobaVerifyingPaymaster (integ test file - boba_aa_sponsoring_fee.test.ts) would be great, if we could add an extra test case to the file covering the validUntil/validAfter case

in case we wanted to add a unit-test, we could add one to ValidateManager.test.ts (modifying paymsterData to include validationData)

packages/boba/bundler/src/modules/ValidationManager.ts
InoMurko: what is the 30 really, magic eh?

packages/boba/bundler/src/modules/ValidationManager.ts
```diff
@@ -162,17 +162,16 @@ export class ValidationManager {
...
+      'expires too soon',
+      ValidationErrors.ExpiresShortly
+    )
+    console.log(`res.aggregatorInfo ${res.aggregatorInfo}`)
+    console.log(res.aggregatorInfo.addr)
+    console.log(res.aggregatorInfo.stake)
+    console.log(res.aggregatorInfo.unstakeDelaySec)
```
InoMurko:
```suggestion
  
```

packages/boba/bundler/src/modules/ValidationManager.ts
souradeep-das: just a buffer it seems 😄

wsdt: true! thanks guys! Added

wsdt:
added couple of things (please see edited changelog on the top of the PR).

packages/boba/bundler/src/modules/ValidationManager.ts
```diff
@@ -162,17 +162,20 @@ export class ValidationManager {
...
+        res.returnInfo.validUntil > (Date.now() / 1000) + 30,
+      'expires too soon',
+      ValidationErrors.ExpiresShortly,
+    )
+    requireCond(
+      res.returnInfo.validAfter == null ||
+        // not adding "buffer" here
+        res.returnInfo.validAfter < Date.now() / 1000,
+      'not valid yet',
+      ValidationErrors.NotValidYet,
```
souradeep-das, line 176: updated the doc with the new error code (https://github.com/bobanetwork/boba/pull/778)

this is better, but just wondering if we should squeeze this into the -32503 error code? (referencing the error codes in the spec https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4337.md)

souradeep-das (approved):
looks great! 🚀

packages/boba/bundler/src/modules/ValidationManager.ts
wsdt: good point! 

Well both are fine for me, advantage would be to stick more to the original spec but disadvantage would be that you don't know if the signature expired or will be valid in future when merging them together (and the error message might be misleading too except we change that altogether). 

`ExpiresShortly = -32503`

InoMurko (approved):
VERY GOOD

PR #779, opened 2023-05-05
add validation of entryPoint and wrapper

### LINKED ISSUE #759: [AA] when bundler starts - verify the correct version of Entrypoint and EntrypointWrapper contract
https://github.com/eth-infinitism/bundler/blob/v0.5.0/packages/bundler/src/BundlerServer.ts#L74-L79

https://github.com/bobanetwork/boba/pull/698/files/ec1d687ded7397d18ff6f1b97cbed506c139c5f7#diff-81644d007bd5f41ea2fc69a9d44934cdd3f84e6a7d482c0de717085729513146R83

this is currently commented out because we can't support custom errors reversals coming from Solidity and our l2Geth,
so this would need to be handled via the wrapper contract, like in ValidationManager...

souradeep-das:
:clipboard: closes #759 

## Overview

Adds validation for EntryPoint and EntryPointWrapper configuration when the bundler starts

## Changes

- make EntryPointWrapper a compulsion
- add validation to confirm EntryPointWrapper points to the EntryPoint contract that is passed
- add validation to confirm EntryPoint contract is valid

## Testing

currently: bundler starts up and integ tests pass


packages/boba/bundler/src/runBundler.ts
```diff
@@ -176,8 +178,19 @@ export async function runBundler(
...
+    entryPointWrapper = epW
+  }
+
+  const entryPointFromWrapper = await entryPointWrapper.entryPoint()
+  if (
+    entryPointFromWrapper.toLowerCase() !== entryPoint.address.toLowerCase()
+  ) {
+    console.error('WARN: entryPointWrapper may be incompatible with entryPoint')
+    process.exit(1)
   }
```
souradeep-das, line 194: EntryPointWrapper has been made a compulsion now

souradeep-das:
currently, no tests for BundlerServer exist, so haven't added an explicit test condition for this misconfiguration case (in which the bundler will not start). lmk what you think.

wsdt (approved):
looks good!

InoMurko (approved):

PR #781, opened 2023-05-05
When bundler starts - verify the correct version of Entrypoint and EntrypointWrapper contract

### LINKED ISSUE #759: [AA] when bundler starts - verify the correct version of Entrypoint and EntrypointWrapper contract

wsdt:
:clipboard: resolves #759

## Overview

Switch to EntryPointWrapper for version check. 

## Changes

- Switch to EntrypointWrapper instead of EntryPoint
- `.addStake()` only available on EntryPoint and doesn't use custom errors, thus kept as is

## Testing

Integration tests.

### DIFF
--- a/packages/boba/bundler/src/BundlerServer.ts
+++ b/packages/boba/bundler/src/BundlerServer.ts
@@ -15,7 +15,7 @@ import { BundlerConfig } from './BundlerConfig'
 import { UserOpMethodHandler } from './UserOpMethodHandler'
 import { Server } from 'http'
 import { RpcError } from './utils'
-import { UserOperationStruct } from '@boba/accountabstraction'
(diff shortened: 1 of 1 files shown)


wsdt:
souradeep is on it

PR #784, opened 2023-05-08
[AA] use EPWrapper instead of senderCreator on aa-sdk

### LINKED ISSUE #777: [AA]: remove the requirement for adding a senderCreatorAddress
simpleAccountAPI on the aa-sdk uses an extra parameter of senderCreatorAddress to compute the counterFactualAddress. Attempt to remove this requirement if this can be fetched from the SimpleAccountFactory instead

souradeep-das:
:clipboard: closes #777 

associated https://github.com/bobanetwork/boba/pull/785

## Overview

Replace the requirement of senderCreator on the aa-sdk with entryPointWrapper. (aa-sdk needs a helper to compute the create2 counterfactual account address)

## Changes

- update entryPointWrapper to catch and return 'getSenderAddress()' from the EP
- update aa-sdk
- add unit tests and update integ tests

## Testing

cd packages/boba/bundler_sdk
yarn test

(and)

integ tests were updated


packages/boba/account-abstraction/contracts/bundler/EntryPointWrapper.sol
```diff
@@ -90,6 +90,11 @@ contract EntryPointWrapper {
         StakeInfo senderInfo, StakeInfo factoryInfo, StakeInfo paymasterInfo,
         AggregatorStakeInfo aggregatorInfo);
 
+    /**
+     * return value of getSenderAddress
+     */
+    error SenderAddressResult(address sender);
```
wsdt, line 96: why do we have a custom error here? thought the whole purpose of this wrapper was to circumvent this?

wsdt (approved):
left a comment

InoMurko (approved):

packages/boba/account-abstraction/contracts/bundler/EntryPointWrapper.sol
souradeep-das: this is just an error type declared to get the selector, custom error is not reverted

PR #785, opened 2023-05-08
[AA-docs] update aa-sdk docs to use entryPointWrapper

souradeep-das:
:clipboard: associated with #784 

## Overview

update aa-sdk docs to use entryPointWrapper instead of senderCreatorAddress

## Changes

/

## Testing

/

### DIFF
--- a/boba_documentation/developer/account-abstraction/bundler-sdk.md
+++ b/boba_documentation/developer/account-abstraction/bundler-sdk.md
@@ -230,8 +230,8 @@ async function wrapProvider(
   originalProvider: JsonRpcProvider, // @ethersproject/providers
   config: ClientConfig,
   originalSigner: Signer = originalProvider.getSigner(), // @ethersproject/abstract-signer
+  entryPointWrapperAddress: string, // must be passed
   wallet?: Wallet, // ethers, must be passed
-  entryPointWrapperAddress?: string // must be passed
 ): Promise<ERC4337EthersProvider>
 ```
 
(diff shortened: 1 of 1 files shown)


boba_documentation/developer/account-abstraction/bundler-sdk.md
```diff
@@ -231,7 +231,7 @@ async function wrapProvider(
   config: ClientConfig,
   originalSigner: Signer = originalProvider.getSigner(), // @ethersproject/abstract-signer
   wallet?: Wallet, // ethers, must be passed
-  senderCreatorAddress?: string // must be passed
+  entryPointWrapperAddress?: string // must be passed
```
wsdt: if it's mandatory, why don't we make it a mandatory function argument? 
Or is this code used somewhere else too where we don't want to break things?

wsdt (approved):
looks good! Left a comment

boba_documentation/developer/account-abstraction/bundler-sdk.md
InoMurko: Tests would need to be adjusted. But doable

InoMurko (approved):

boba_documentation/developer/account-abstraction/bundler-sdk.md
souradeep-das: its optional for the low-level API, but mandatory for the high-level API, updated the other PR with this change

PR #793, opened 2023-05-10
Add runtime flag to enable debug methods

### LINKED ISSUE #761: [AA] debug_ namespace should not be available
`BundlerServer.ts`
```
// case 'debug_bundler_clearState':
      //   this.debugHandler.clearState()
      //   result = 'ok'
      //   break
      // case 'debug_bundler_dumpMempool':
      //   result = await this.debugHandler.dumpMempool()
      //   break
      // case 'debug_bundler_setReputation':
      //   await this.debugHandler.setReputation(params[0])
      //   result = 'ok'
      //   break
      // case 'debug_bundler_dumpReputation':
      //   result = await this.debugHandler.dumpReputation()
      //   break
      // case 'debug_bundler_setBundlingMode':
      //   await this.debugHandler.setBundlingMode(params[0])
      //   result = 'ok'
      //   break
      // case 'debug_bundler_setBundleInterval':
      //   await this.debugHandler.setBundleInterval(params[0], params[1])
      //   result = 'ok'
      //   break
      // case 'debug_bundler_sendBundleNow':
      //   result = await this.debugHandler.sendBundleNow()
      //   if (result == null) {
      //     result = 'ok'
      //   }
      //   break
```

These seem useful (locally) but should not be available on testnets and prod mainnet environments. 
These should be enabled/disabled with runtime flags.

wsdt:
:clipboard: resolves #761

## Overview

Add runtime flag for enabling/disabling debug methods.

## Changes

- Adding runtime flag
- Modifying BundlerServer to react to runtime flag

## Testing

Test run & tests.


packages/boba/bundler/src/runBundler.ts
```diff
@@ -226,7 +227,8 @@ export async function runBundler(
     debugHandler,
     config,
     provider,
-    wallet
+    wallet,
+    programOpts.enableDebugMethods ?? false
```
souradeep-das: the opt would be parsed into config, so config could be used for getting the `enableDebugMethods` too, but this is fine as well!

(just in case we do that - we could add the default value to false on the BundlerConfig as well)

packages/boba/bundler/src/runBundler.ts
```diff
@@ -119,6 +119,7 @@ export async function runBundler(
     .option('--addressManager <string>', 'address of the Address Manager', '')
     .option('--l1NodeWeb3Url <string>', 'L1 network url for Address Manager', '')
     .option('--maxBundleGas <number>', 'Max Bundle Gas available to use', '5000000')
+    .option('--enableDebugMethods', 'debug_* methods available', false)
 
   const programOpts = program.parse(argv).opts()
   showStackTraces = programOpts.showStackTraces
```
souradeep-das, line 125: could we log a warning - if debug mode is enabled? (helps to know the configuration)

souradeep-das (approved):
lgtm!

packages/boba/bundler/src/runBundler.ts
wsdt: added!

InoMurko (approved):

PR #799, opened 2023-05-10
feat: add forcePush duration on message relayers

souradeep-das:
:clipboard:

## Overview

Currently, the message relayer does not relay messages when the gas price is over the configured max gas price.
This adds a maximum waiting time after which the relayers will relay message at whatever is the current network gas price

## Changes
- adds FORCE_PUSH_PERIOD_S, seconds to wait before force pushing. If a message has been waiting for challengePeriod + FORCE_PUSH_PERIOD_S, it will be forced
- adds a highGasPriceInGwei setting, to set a very high gas price, in which case the relayer will pause (to account for short term spikes)
- adds logic for forcePush for both mr and mrf
- moves the gasPriceAccepted check after cleaning up the subbuffer to account for precise checking of the forcePeriod

## Testing

tested on local stack, needs more testing


packages/message-relayer/src/service.ts
```diff
@@ -241,32 +250,34 @@ export class MessageRelayerService extends BaseService<MessageRelayerOptions> {
           (bufferFull || timeOut) &&
           pendingTXTimeOut
         ) {
-          if (gasPriceAcceptable) {
```
souradeep-das, line 244: I have moved this check a bit below - for precise checking of the first item in the buffer - and wether it should be forced. (sliced buffers)

The effect is depending on the MAX_WAIT_TX_TIME_S set, the relayer could work a bit more when it comes to checking if a message is relayed and check its buffer is valid

packages/message-relayer/src/service.ts
```diff
@@ -241,32 +250,34 @@ export class MessageRelayerService extends BaseService<MessageRelayerOptions> {
...
+            ) {
+              newMB.push(cur)
             }
-            this.state.messageBuffer = newMB
-
+          }
+          this.state.messageBuffer = newMB
+          if (
+            gasPriceAcceptable ||
+            (await this._forcePush(this.state.messageBuffer)) === true
```
wsdt, line 278: could make sense to add a test for `forcePush`?

wsdt (approved):
left a comment :-)

boyuan-chen (approved):

InoMurko:
what's the status on this?

souradeep-das:
this is ready to be merged. but which one should we pick?

If we still intend to have the functionality to force messages out of the relayer we could merge this. We will need to run this on the testnet for a while though, to fully check
If the gas prices have now gone down, and this is not an immediate problem and to avoid testing this - we could keep the PR open until relayers are removed

InoMurko:
What do you mean with `but which one should we pick?`?
Relayers will get removed eventually - the timeline is about 2 months from now.

souradeep-das:
what should we do?
If this isn't an issue anymore (or we expect to not encounter the associated issue in the next 2 months), I would say lets keep this open for now? (so as to avoid testing and updating to this new feature)

PR #806, opened 2023-05-11
Add exit dumper

boyuan-chen:

## Overview

This PR is for the migration.

Since L2ToL1MessagePasser.sol is updated in the v3 and it's behind the proxy, we need to dump the exit message and write them to new storage to avoid the double exit in the v3.

## Changes

- Add dumper struct and code to write data to a file

## Testing

The unit tests are added

### DIFF
--- a/l2geth/core/vm/evm.go
+++ b/l2geth/core/vm/evm.go
@@ -34,12 +34,16 @@ import (
 	"github.com/ethereum-optimism/optimism/l2geth/rollup/rcfg"
 	"github.com/ethereum-optimism/optimism/l2geth/rollup/util"
 	"github.com/ethereum-optimism/optimism/l2geth/rpc"
+	"github.com/ethereum-optimism/optimism/l2geth/statedumper"
 	"golang.org/x/crypto/sha3"
 )
 
 // emptyCodeHash is used by create to ensure deployment is disallowed to already
 // deployed contract addresses (relevant after the account abstraction).
 var emptyCodeHash = crypto.Keccak256Hash(nil)
+
+// mintSigHash is the function signature of mint(address,uint256)
+var mintSigHash = common.FromHex("0x40c10f19")
 
 type (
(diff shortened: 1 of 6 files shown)


wsdt (approved):

InoMurko (approved):

boyuan-chen:
I bumped up go version to v1.18 so that it supports `os.CreateTemp` and `os.ReadAll`.

l2geth/core/vm/evm.go
```diff
@@ -579,6 +583,20 @@ func (evm *EVM) bobaTuringCall(input []byte, caller common.Address, mayBlock boo
 // the necessary steps to create accounts and reverses the state in case of an
 // execution error or failed value transfer.
 func (evm *EVM) Call(caller ContractRef, addr common.Address, input []byte, gas uint64, value *big.Int) (ret []byte, leftOverGas uint64, err error) {
+	if addr == dump.MessagePasserAddress {
+		statedumper.WriteMessage(caller.Address(), input)
+	}
+
+	if addr == dump.OvmEthAddress {
```
souradeep-das, line 590: what is the purpose of this?

is this for getting ETH holders? if then there could be other recipients of ETH through the native transfers?

souradeep-das (commented):
lgtm! 
but should we merge this to develop? given this is to be run on the replica?

l2geth/core/vm/evm.go
boyuan-chen: This is another security layer to make sure that we get the ETH holder addresses that are used to move the ETH balance from OVM_ETH contract to the standard state in the db.

boyuan-chen: All ETH holder addresses should be fetched via: https://github.com/bobanetwork/v3-anchorage/blob/boba-chain-ops/boba-chain-ops/cmd/eth-crawler/main.go

souradeep-das: cool!

PR #813, opened 2023-05-12
Add more information to Bundler SDK release & versioning

wsdt:
## Overview

`npm version` is not persisted, which might break second release. 
Add more information to the npm package for npmjs.org. 

## Changes

- npm version executed on PR commenter and committed to current branch. 
- Add more information to package json of Bundler SDK for npmjs.org

## Testing

On PR tested


packages/boba/bundler_sdk/package.json
```diff
@@ -1,7 +1,20 @@
 {
   "name": "@bobanetwork/bundler_sdk",
-  "version": "0.2.3",
+  "version": "0.3.0",
```
wsdt, line 3: this is needed, to avoid conflict in the next release cycle as 3.0 as already been released

.github/workflows/aa-bundler-sdk.yml
```diff
@@ -19,11 +19,27 @@ jobs:
...
+        run: |
+          git config user.name "GitHub Actions Bot"
+          git config user.email "<>"
+
+      - name: commit
+        run: |
+          # Stage the file, commit and push
+          git add packages/boba/bundler_sdk/package.json yarn.lock
+          git commit -am "Bump Bundler SDK version"
+          git push
```
souradeep-das: nice! 
but i'm thinking - does this work on the develop branch? it should have branch protections and reject direct push?

wsdt: yeah thought about this too, but adding this commit/push to the Pr-commenter causes many more issues, such as bumping the version on every commit or only on the first commit, etc 

That would be the cleanest and my hope is that the Github token bypasses this (otherwise it could be something TBD if this action can possibly bypass the branch protection since it really just adds the package json and yarn.lock.

souradeep-das: hmm makes sense. lets see what permissions the bot has and the rules for the branch. if it is an admin, maybe it can force!
It would be fantastic if there was a way to bump once for a pr effectively, or if there was a github action plugin that did it 😄

wsdt: @InoMurko can we try this? The only other alternative I see would be creating a 3rd Github action that is triggered on Pull-Request close (and hoping that the new commit will be merged along with it).

InoMurko: is this meant to be executed on `develop`? 

we currently have releases coming from develop -> master, and all releases are coming from there. does this go with this flow?

wsdt: yes the npm version is meant to be executed in `develop`. Basically on push when something in the bundler_sdk or utils has changed.

InoMurko: that doesn't sound right - develop is not for releases and could be unstable.

InoMurko: I would prefer releases go out of master - like everything else. check the release flow we currently have.

souradeep-das (approved):
nice!

.github/workflows/aa-bundler-sdk-pr.yml
```diff
@@ -1,19 +1,26 @@
 name: Will trigger release of AA-Bundler to NPM (PR comment)
 on:
   pull_request:
+    branches:
+      - master
+
+permissions:
+  contents: write
```
InoMurko: should this be?
```suggestion
name: Will trigger release of AA-Bundler to NPM (PR comment)
on:
  push:
    tags:
    - '*'

permissions:
  contents: write
```

.github/workflows/aa-bundler-sdk-pr.yml
```diff
@@ -1,8 +1,8 @@
 name: Will trigger release of AA-Bundler to NPM (PR comment)
 on:
-  pull_request:
-    branches:
-      - master
+  push:
+    tags:
+    - '*'
```
souradeep-das: wondering if there's an effect, when the PR is merged. (push to develop)
if this workflow runs again is there an impact to the patch version?

InoMurko: push + tags means this workflow would only run if there is a tag pushed to git.
which in our approach is when we make a release on master and add a tag to it.
but since this does a git push
```
- name: commit
       if: steps.version-updated.outputs.has-updated == 'false'
       run: |
         # Stage the file, commit and push
         git add packages/boba/bundler_sdk/package.json yarn.lock
         git commit -am "Bump Bundler SDK version"
         git push
         # $GITHUB_HEAD_REF
```
I'm uncertain how this will work - we should try.

wsdt: yeah don't think this will work then - what's wrong with merge to master? 

Edit: Doesn't work in the sense of the version change being included in the new tag. But it should work with the npm release itself.

wsdt: > wondering if there's an effect, when the PR is merged. (push to develop) if this workflow runs again is there an impact to the patch version?

No, if the changeset includes already a version change for the bundler sdk then it won't run again :-)

wsdt: > wondering if there's an effect, when the PR is merged. (push to develop) if this workflow runs again is there an impact to the patch version?

No, if the changeset includes already a version change for the bundler sdk then it won't run again :-)

InoMurko (approved):

InoMurko:
will take another look tomorrow if it fits our needs

Issue #795, opened 2023-05-10
Upgrade Teleportation (contract) for variety of tokens and native assets


InoMurko:
## Issue Type

```
[ ] bug report
[ x] feature request
```

## Current Behavior
Currently Teleportation only works with BOBA as Native asset on (BNB, Avax BOBA) and ERC20 on Ethereum Boba.

## Expected Behavior
We would like the Teleportation to support variety of assets (ERC20 and native assets).

**Freely Teleport BOBA between**: (this should already work)
Ethereum > Ethereum Boba
Ethereum > Avax Boba
Ethereum > BNB Boba

Ethereum Boba > Ethereum
Avax Boba > Ethereum
BNB Boba > Ethereum

Ethereum Boba > Avax Boba
Avax Boba > BNB Boba
BNB Boba > Avax Boba

Ethereum > BNB Boba

**Freely Teleport ETH between**:
Ethereum > Ethereum Boba
Ethereum Boba > Ethereum
(eth doesn't make sense on other chains)

**Freely Teleport USDT between**:
Ethereum > Ethereum Boba 
Ethereum Boba  > Ethereum

**Freely Teleport BNB between**:
Ethereum > Ethereum Boba 
Ethereum Boba  > Ethereum

MVP STATE:
=========
When bridging ETH:
ETH L1 <> BobaETH
BobaAvax > BobaETH
BobaBNB > BobaETH
BobaFantom > BobaETH

When bridging BOBA:
ETH L1 <> All Boba Chains
BobaAvax <> ETH L1 + All Boba Chains
BobaBNB <> ETH L1 + All Boba Chains
BobaFantom <> ETH L1 + All Boba Chains

InoMurko:
<img width="358" alt="image" src="https://github.com/bobanetwork/boba/assets/2582555/ce0c0bf3-b335-48fd-966e-dfa324dc1b39">

souradeep-das:
So we want Teleportation routes even for routes where classic bridging is possible right? to offer a cheaper way to bridge?
(for instance bridging ETH, ETH L1 <> BobaETH)

wsdt:
[ ] Request audit(s) for diff afterwards

wsdt:
@InoMurko @cryptoxman How do we want to handle daily limit, min/max deposit, etc? 
Just handle them separately per asset?

PR #822, opened 2023-05-15
Switch from file-based db to postgresql with DB migrations

### LINKED ISSUE #791: Teleportr database
Use PG for database and drop file based database

## Issue Type

```
[ ] bug report
[ x] feature request
```

## Current Behavior
```
  const DATABASE_PATH = config.str(
    'database-path',
    env.DATABASE_PATH || '../db'
  )
```

## Expected Behavior
Use Postgres

wsdt:
:clipboard: resolves #791

## Overview

Switch from file-based db to postgresql in Teleporter and support db migrations. 

## Changes

- Adding teleportation_db to docker-compose-side.yml
- Adding corresponding .env variables
- Modification of interfaces, etc.
- typeorm integration

## Testing

Project build & tests ran.


wsdt:
4 tests failing, rest succeeding

wsdt:
Check before merging: 
https://github.com/bobanetwork/boba/issues/846

wsdt:
Tests succeeding, will now have a look at https://github.com/bobanetwork/boba/issues/846

DO NOT MERGE YET

wsdt:
fixing coverage -> moving over to integration tests
--> hopefully minor issue open (in progress)

integration-tests/test/eth-l2/teleportation.spec.ts
```diff
@@ -10,21 +10,21 @@ import {
...
+import { ChainInfo } from '@boba/teleportation/src/utils/types'
 
 /* Imports: Core */
-import { TeleportationService } from '../dist/service'
+import { TeleportationService } from '@boba/teleportation/src/service'
+import { AppDataSource, historyDataRepository } from "@boba/teleportation/src/data-source";
+import { HistoryData } from '@boba/teleportation/src/entity/HistoryData'
 
-describe('teleportation', () => {
+describe.only('teleportation', () => {
```
InoMurko: what does this only do?

packages/boba/teleportation/package.json
```diff
@@ -1,43 +1,50 @@
...
-    "@eth-optimism/common-ts": "0.2.2",
-    "@eth-optimism/contracts": "0.5.11",
-    "bcfg": "^0.1.6",
-    "dotenv": "^8.2.0",
-    "ethers": "^5.5.4",
-    "lodash": "^4.17.21"
-  }
+   "name": "@boba/teleportation",
+   "version": "1.0.0",
+   "description": "The teleportation between L2s",
```
InoMurko:
```suggestion
   "description": "The teleportation between L2s and L1s",
```

packages/boba/teleportation/.env.example
```diff
@@ -3,4 +3,8 @@ TELEPORTATION_DISBURSER_KEY=
 # Optional
 POLLING_INTERVAL=
 BLOCK_RANGE_PER_POLLING=
-DATABASE_PATH=
+POSTGRES_PASSWORD=abcdef
+POSTGRES_DB_HOST=teleportation_db
+POSTGRES_DB=postgres
+POSTGRES_PORT=5432
+POSTGRES_USER=postgres
```
InoMurko: this is a nice to have, but this reminds me of prefixing env vars for where they belong:
```
TELEPORTATION_POSTGRES_PASSWORD=abcdef
TELEPORTATION_POSTGRES_DB_HOST=teleportation_db
TELEPORTATION_POSTGRES_DB=postgres
TELEPORTATION_POSTGRES_PORT=5432
TELEPORTATION_POSTGRES_USER=postgres
TELEPORTATION_POLLING_INTERVAL=
TELEPORTATION_BLOCK_RANGE_PER_POLLING=
```
sometimes env vars get mixed or all secrets end up in one bucket and then they collide

InoMurko (approved):
lgtm - 

I think the only thing that *needs* fixing is 
```
describe.only('teleportation', () => {
```
not sure if the whole test gets executed with this only

integration-tests/test/eth-l2/teleportation.spec.ts
wsdt: The PR is not ready to merge, the last thing missing is to get the new integration test talking to the postgres db.
discovered that after the test coverage failed to move the test over to integration tests.

packages/boba/teleportation/.env.example
wsdt: sure will do!

integration-tests/test/eth-l2/teleportation.spec.ts
InoMurko: ah ok! I will then turn the PR into draft again and remove the reviewers as we get emails and notifications for commits

wsdt: thx!

wsdt:
Integration tests seem to & should pass now

integration-tests/test/eth-l2/teleportation.spec.ts
wsdt, line 1: now separately added as chain specific configs are otherwise not taken and despite the correct command in the dockerfile CircleCI seems to just execute the tests on boba eth. 

Sticking to what we have been doing for other tests too.

integration-tests/test/eth-l2/erc1155_bridge.spec.ts
```diff
@@ -1128,7 +1128,7 @@ describe('ERC1155 Bridge Test', async () => {
 
       const approveTX = await L1ERC1155.connect(
         env.l2Wallet_2
-      ).setApprovalForAll(L1Bridge.address, true)
+      ).setApprovalForAll(L1Bridge.address, true, { gasLimit: 1000000 })
```
wsdt, line 1131: failing test (only fails on boba eth), comparing alt test showed manual gas limit should fix the test

.github/workflows/ts-packages.yml
```diff
@@ -126,7 +126,8 @@ jobs:
 
       - name: Check packages/boba/teleportation
         working-directory: ./packages/boba/teleportation
-        run: npx depcheck
+        # needed by typeorm
+        run: npx depcheck --ignores @types/node
```
InoMurko: there's ```.depcheckrc``` for this

integration-tests/test/alt-l2/teleportation.spec.ts
```diff
@@ -535,4 +541,43 @@ describe('teleportation', () => {
       expect(preBOBABalance.sub(postBOBABalance)).to.be.eq(0)
     })
   })
+
+  describe('database connection', () => {
```
InoMurko: this is not an integration test but a unit test.

InoMurko (changes requested):

integration-tests/test/alt-l2/teleportation.spec.ts
wsdt: database connection? How is that supposed to be a unit test? 
I added these tests specifically to see if there is an actual connection

integration-tests/test/eth-l2/erc1155_bridge.spec.ts
```diff
@@ -1116,7 +1116,7 @@ describe('ERC1155 Bridge Test', async () => {
       expect(postL1Balance).to.deep.eq(preL1Balance.add(DUMMY_TOKEN_AMOUNT_3))
     })
 
-    it('should be able to deposit token to another wallet on L2', async () => {
+    it('should be able to deposit token to another wallet on L2 (2)', async () => {
```
InoMurko, line 1119: what exactly is the purpose of this test compared to the one ```
should be able to deposit token to another wallet on L2 (1)
```
?

**Review by InoMurko — CHANGES_REQUESTED** [review]

File: integration-tests/test/eth-l2/erc1155_bridge.spec.ts (hunk 7 above)
Comment (wsdt) [reply integration-tests/test/eth-l2/erc1155_bridge.spec.ts:1119]: not sure, because I didn't write them. 

Since our tests depend on each other, i assume these are there to ensure that the state from previous tests is correct and the deposit for those two different wallets work (maybe different roles/privileges)

File: integration-tests/test/alt-l2/teleportation.spec.ts (hunk 6 above)
Comment (InoMurko) [reply integration-tests/test/alt-l2/teleportation.spec.ts]: the reason for this being a bad integration test is that you don't have an outside view of the application, but an inside view of the application.
so, you're using internal data structures to test whether postgres is consuming your data.

an integration test would be possible if:
teleportation as a service would expose an API that allows you to query the latest deposit.
you make a deposit, via geth and solidity, check if teleportation matches your assumption about the deposit.
in your case, you actually brought DB schema to the integraiton test that's on the parent monorepo.
I think this is too much coupling.

also, like - wouldn't other tests already cover whether you have a connection or not? if you make several deposits one after another - if blocks are not in DB - how would the application proceed to recognise new deposits for a chainid?

File: integration-tests/test/eth-l2/erc1155_bridge.spec.ts (hunk 7 above)
Comment (InoMurko) [reply integration-tests/test/eth-l2/erc1155_bridge.spec.ts:1119]: please take a look then.
also tests should be decoupled, if they're coupled make an issue so that they get fixed.

File: integration-tests/test/alt-l2/teleportation.spec.ts (hunk 6 above)
Comment (wsdt) [reply integration-tests/test/alt-l2/teleportation.spec.ts]: yes other tests would fail, tbh I added them because I they helped me discovering a typing issue with the db. 
But will remove!

**Review by InoMurko — APPROVED** [review]

File: integration-tests/test/eth-l2/erc1155_bridge.spec.ts (hunk 7 above)
Comment (wsdt) [reply integration-tests/test/eth-l2/erc1155_bridge.spec.ts:1119]: the 2nd test is about depositing from another wallet (that's why we call `connect()` and have to call approve before. 
So yes it actually makes sense to have both. 

Especially since the 2nd still fails only on Boba eth for some reason. 

And yes, one needs to execute the whole describe for the tests to succeed as they depend on each other in general

File: packages/boba/teleportation/src/service.ts [hunk 8]
```diff
@@ -295,32 +294,25 @@ export class TeleportationService extends BaseService<TeleportationOptions> {
...
-      const addrsPath = path.resolve(dumpsPath, `depositInfo-${chainId}.json`)
-      await fsPromise.writeFile(addrsPath, JSON.stringify({ latestBlock }))
+      const historyData = new HistoryData()
+      historyData.chainId = chainId
+      historyData.blockNo = latestBlock
+      await historyDataRepository.save(historyData)
     } catch (error) {
       this.logger.error(`Failed to put depositInfo! - ${error}`)
     }
   }
```
InoMurko, line 305: this try - catch seems weird.
imo a try catch is such an obscure part of any language, way too much usage.
```
const saved = await historyDataRepository.save(historyData)
if saved === false {
  this.logger.error(`Failed to put depositInfo! - ${error}`)
}
```

what can actually fail in this block of code? creating the object or saving into the DB?


```
try {
      lastBlock = await this._getDepositInfo(chainId)
    } catch (e) {
      this.logger.warn(`No deposit info found in chainId - ${chainId}`)
      lastBlock = depositTeleportation.height
      // store the new deposit info
      await this._putDepositInfo(chainId, lastBlock)
    }
```

this one too. try catch is for **exception** handling, not normal code branches.
I know you didn't write this, just laying it out there for future reference. don't use try catch or minimise it's usage

InoMurko: My comment is wrong is the DB insertion actually throws exceptions. But then we don't need two wrapped try catches either... :-)

wsdt: > this try - catch seems weird. imo a try catch is such an obscure part of any language, way too much usage.
> 
> ```
> const saved = await historyDataRepository.save(historyData)
> if saved === false {
>   this.logger.error(`Failed to put depositInfo! - ${error}`)
> }
> ```
> 
> what can actually fail in this block of code? creating the object or saving into the DB?
> 
> ```
> try {
>       lastBlock = await this._getDepositInfo(chainId)
>     } catch (e) {
>       this.logger.warn(`No deposit info found in chainId - ${chainId}`)
>       lastBlock = depositTeleportation.height
>       // store the new deposit info
>       await this._putDepositInfo(chainId, lastBlock)
>     }
> ```
> 
> this one too. try catch is for **exception** handling, not normal code branches. I know you didn't write this, just laying it out there for future reference. don't use try catch or minimise it's usage

Just replicated the prior code to not move away from current behavior. 
But happy to adapt it!

wsdt:
There is one seemingly unrelated integration test named `should be able to deposit token to another wallet on L2 (2)` that seems to fail deterministically: 
https://app.circleci.com/pipelines/github/bobanetwork/boba/3371/workflows/1bbf7589-6edc-452b-b26c-a2753e3756f5/jobs/13339


Although, locally when executing the test file all tests including the one above explicitly run through successfully, the same applies when only executing the inner describe - all tests including the one above run through. So possibly it also has something to do with how CircleCI is setup, as I was running the integration tests within Docker as well. 

They also run through on Circle CI when isolating them: 
https://app.circleci.com/pipelines/github/bobanetwork/boba/3680/workflows/a375fe31-0813-4302-b016-af006704374c/jobs/13340

https://app.circleci.com/pipelines/github/bobanetwork/boba/3681/workflows/9d5bb686-cb3d-4912-87f9-ffae249a8b83/jobs/13347

BNB and AVAX tests run through too - the same test (1:1 code) runs through on Alt-L2 while failing on Boba Eth. 
https://app.circleci.com/pipelines/github/bobanetwork/boba/3371/workflows/1bbf7589-6edc-452b-b26c-a2753e3756f5/jobs/12124
https://app.circleci.com/pipelines/github/bobanetwork/boba/3371/workflows/1bbf7589-6edc-452b-b26c-a2753e3756f5/jobs/12125

It seems like that completely unrelated integration tests in other integration test files make that single test fail. 

As all teleporter integration tests run through successfully, as so do other integration tests - I suggest that we look at our integration tests separately as it really looks like that we should have a look at the dependencies between our integration tests. 

Feels a bit out of scope for this and I depend on this code for continuing with: 
https://github.com/bobanetwork/boba/issues/795

@InoMurko @souradeep-das please let me know what you think. 

Created an issue and would treat it as independent issue: 
https://github.com/bobanetwork/boba/issues/953

Interestingly, this PR only replaces the file based storage in the teleporter with a postgresql. Which shouldn't have any impact on any integration tests besides teleporter tests & tests being dependent on the teleporter as long as the postgres db is running.

wsdt:
Created an issue: 
https://github.com/bobanetwork/boba/issues/953

packages/boba/teleportation/src/service.ts
InoMurko: You changed the underlying database storage. You should change the current behaviour, especially if it's wrong and/or harmful.

InoMurko:
```
There is one seemingly unrelated integration test named should be able to deposit token to another wallet on L2 (2) that seems to fail deterministically:
```
deterministically or probalisticly?

InoMurko (approved):
lets see how flaky things get in the main repo

souradeep-das:
yeah its weird, sometimes we have seen it is because of the GasOracle not being able to estimateGas precisely and that fails the tests. Passing an explicit { gasLimit: 7000000 } to the transaction mostly makes it back to working. (but not sure in this case)
but I assume you have already tried doing that, since you added that on the other test?

souradeep-das (approved):
looks great to me!

wsdt:
Gateway dependencies broke the project/got into conflict with the added dependencies in this PR. 
Works now again

Issue #828, opened 2023-05-16
dtl making huge range requests on L1


InoMurko:
## Issue Type

```
[ x] bug report
[ ] feature request
```

## Current Behavior
Log from L2 Boba Goerli
`
[...]

PR #835, opened 2023-05-18
Verifier fixes

mmontour1306:

## Overview

Updates to run boba_community/fraud-detector against Mainnet and Goerli networks.

## Changes

- Pins web3.py to a working version, before a breaking API change
- Edits docker-compose.yml files to better match the Replica ones in boba_node/...
- Adds a DTL workaround to avoid excessive range queries for AddressManager events

## Testing

Run a local instance of the Verifier + Fraud Detector and see if it's able to sync to the tip of Mainnet and Goerli.

NOTE - most of the changes will only affect the Verifier but the DTL change is in common code which will also affect Replica and Sequencer nodes. May fix issue #828.


InoMurko (approved):

packages/data-transport-layer/src/services/l1-ingestion/service.ts
```diff
@@ -485,6 +486,12 @@ export class L1IngestionService extends BaseService<L1IngestionServiceOptions> {
         this.state.startingL1BlockNumber,
         blockNumber
       )
+    } else if (contractName in this.state.addrList) {
+      /* Workaround to avoid expensive queries for deployments where
+         the contract address has not changed, and the addrList value
```
souradeep-das, line 491: could there be a case where the contract address changed and was re-registered?

mmontour1306: Theoretically possible but it hasn't happened on our mainnet. I don't know about the alt-L1s. When we did our V1->V2 regenesis the contract names also changed (old ones had an OVM_ prefix) so there was no address change.

boyuan-chen (approved):

PR #837, opened 2023-05-18
[AA] First draft defaults

### LINKED ISSUE #760: [AA] consider good defaults for our networks
BundlerConfig.ts has a few hardcoded configs for now:
```
// TODO: consider if we want any default fields at all
// TODO: implement merging config (args -> config.js -> default) and runtime shape validation
export const bundlerConfigDefault: Partial<BundlerConfig> = {
  port: '3000',
  entryPoint: '0x1306b01bC3e4AD202612D3843387e94737673F53',
  unsafe: false,
  conditionalRpc: false,
  minStake: '1',
  minUnstakeDelay: 60,
  autoBundleInterval: 1,
  autoBundleMempoolSize: 1,
}
```

we should consider good default configuration going forward (especially minStake, minUnstakeDelay, autoBundleInterval, autoBundleMempoolSize)

wsdt:
:clipboard: Resolves #760

## Overview

Defaults for AA bundler

## Changes

- Defaults taken by eth-infinitism (although other implementations don't seem to define defaults for these params at all)


souradeep-das:
nice! I think we should have Bundler Defaults. the configurations are merged in the following order (defaults -> config.js -> shell args) with shell args having the precedence when conflicts

In our case, we don't use the config.js file, so its just defaults + shell args
and all of the non-optional params are either set through defaults or shell args except gasFactor which is set to a default on runBundler.js. 
a) So we could add gasFactor to the defaults

b) (i'm not sure about this one) entryPoint could be made "optional" like the entryPointWrapper since we are fetching it from addressManager and removed from the defaults. But otherwise we could set the default value for entryPoint to "0xa6e2cbb294d3b84e7900daf0052ffe26bb1328ff" (this is the addr with the default depl proxy)

c) minStake - context: The stake on the EntryPoint is never slashed but the stake exists to prevent sybill/dos attacks. Accounts (or Paymasters) only need to be staked if they have > 4 entries on the mempool. or a paymaster that access its own storage and there are storage collisions because of concurrent requests and validation fails
Which means for main networks we should set the stake sufficiently high. 

I think a way to determine a value of this could be:
min Stake amount = (min_attack_cost * 4) / bundler_requests_can_sustain  

approximating 1000 reqs with 50 ETH cost of attack, how about 0.2 ETH for mainnet? and 0.0001 ETH for goerli? (or 1 should not be a problem too)
so defaults can be set to that of georli's, while we pass minStake as a shell parameter on bundler.sh

d) minUnstakeDelay for the same reason, default of 0 sounds good. maybe 0/ 5mins for goerli and 1 day on mainnet?
since we set this explicitly on mainnet, lets pass this as a shell parameter on bundler.sh too?

e) autoBundleInterval and autoBundleMempoolSize defaults look good to me. In case we want to update this for mainnet and remember about it, lets pass this on bundler.sh too

souradeep-das:
also added a doc here https://github.com/bobanetwork/boba/pull/842

wsdt:
done @souradeep-das, thanks!

souradeep-das (approved):
looks great! 
(seems like the last config change makes the bundler throw an error)

wsdt:
> looks great! (seems like the last config change makes the bundler throw an error)

fixed! thx

InoMurko (approved):

PR #845, opened 2023-05-24
Adding storybook support

alvaro-ricotta:
We have updated the version of React Scripts to version 5.0, which is a necessary requirement to start using Storybook.

Since the new versions do not come with polyfills by default, we manually installed the polyfills for the following libraries:

1. process
2. stream
3. fs
4. path
5. assert
6. http
7. https
8. os
9. zlib
10. buffer

Finally, after updating React Scripts, we were able to proceed with the installation of Storybook. It is now functional using the command :

`"yarn storybook".`

I have installed the option to use the Light/Dark theme by default with the new styled components methodology. I also included a version to preview both themes at the same time.

New components have been sent:

- Button
- InputMax
- Switch Button.
- Svg Color Switcher
- Dropdown Form

Additionally, I have defined the color palette for both themes.

Furthermore, the corresponding configuration for the required typography has been added, including titles and regular texts.


sk-enya:
@alvaro-ricotta 
here is my findings.
1.  there is conflict.
2. unable to run the `yarn install`  it's failing with random exceptions for node version.
3. unable to move further with `yarn start & yarn storybook`

Would be great if you can add more details like which version of node have to use.

ops_boba/api/banxa-api/serverless-banxa.yml
```diff
@@ -0,0 +1,23 @@
+service: banxaintegration
```
sk-enya: Why do we have banxa changes here?  are you sure this is working correctly?

alvaro-ricotta: Nice catch!, I already remove it.

sk-enya (approved):
All Looks good!

Issue #843, opened 2023-05-23
[AA] fix: bundler relies on eth_getLogs with indefinite range


souradeep-das:
## Issue Type

```
[X] bug report
[ ] feature request
```

## Current Behavior
L2 does not support eth_getLogs with ranges more than 5000. As a result bundler fails to spin up for test/main networks

error could be of the form
```
/opt/optimism/packages/boba/bundler/bundler.js:2
...
```

or 

```
{"jsonrpc":"2.0","error":{"code":-32000,"message":"exceed maximum block range: 5000"},"id":48}\n',
  error: Error: exceed maximum block range: 5000
```

The reason for the error is here: https://github.com/bobanetwork/boba/blob/develop/packages/boba/bundler/src/modules/EventsManager.ts#L42

## Expected Behavior
Bundler starts and runs for networks with eth_getLogs limits

## Steps to Reproduce
- change L2_NODE_WEB3_URL on the bundler to point to the test network
- try spinning up the bundler

## Suggested Fix
Find a way to accomodate the bundler requests within the specified 5k range

Can use: a local memory counter for the current read counter
and for when it starts the first time/ restarts split it into 5000 block requests

vortex-hue:
hi @souradeep-das , please can I get assigned to this task, and I'd get to work on it as soon as possible.

PR #855, opened 2023-05-26
fix: bundler relies on eth_getLogs with indefinite range

souradeep-das:
:clipboard: https://github.com/bobanetwork/boba/issues/843 #866 

## Overview

Fixes bundler reliance on eth_getLogs with range > 5000

## Changes

- bundler splits eth_getLogs requests if the range to request is > 5k
- a new config 'l1Offset' has been added to bundler, to allow the bundler to start watching from a specific block
- adds opts {fromBlock and toBlock} for eth_getUserOperationReceipt and eth_getUserOperationByHash

## Testing

circleci logs


packages/boba/bundler/src/modules/EventsManager.ts
```diff
@@ -40,9 +46,26 @@ export class EventsManager {
...
    */
   async handlePastEvents(): Promise<void> {
+    let startBlock = this.lastBlock
+    const currentBlock = await this.provider.getBlockNumber()
+    if (currentBlock > startBlock + 5000) {
+      while (startBlock + 5000 < currentBlock) {
+        const events = await this.entryPoint.queryFilter(
+          { address: this.entryPoint.address },
+          startBlock,
+          startBlock+4999
```
souradeep-das: assuming fromBlock and toBlock in eth_getLogs range is inclusive (need to double check)

packages/boba/bundler/src/BundlerServer.ts
```diff
@@ -186,10 +186,10 @@ export class BundlerServer {
         )
         break
       case 'eth_getUserOperationReceipt':
-        result = await this.methodHandler.getUserOperationReceipt(params[0])
+        result = await this.methodHandler.getUserOperationReceipt(params[0], params[1])
         break
       case 'eth_getUserOperationByHash':
-        result = await this.methodHandler.getUserOperationByHash(params[0])
+        result = await this.methodHandler.getUserOperationByHash(params[0], params[1])
```
InoMurko, line 192: is there a specification attached to eip-4337 and is this breaking it?

packages/boba/bundler/src/modules/EventsManager.ts
InoMurko: should we extract away the 4999? could be less, could be more!

InoMurko (changes requested):

packages/boba/bundler/src/BundlerServer.ts
souradeep-das: hmmm right, the original specs specify one parameter. 
The second parameter here is optional though, but do we want to stick with one param and split into multiple requests?

wsdt: if we go with multiple request we could think about `Promise.all`

InoMurko: as you said @souradeep-das the additional parameters are optional so things don't break the specs.
whereas I'm now doubting the quality or usability of the specs.

the client either needs to keep track of *when* (in terms of blocks) a user operation occurred so that it supplies the optional params, or as per specs, scan ALL blocks, not all providers allow you to do that.

@souradeep-das would this chunk of code scan all logs from the deployment of `entryPoint` till `latest` block?
```
const event = await this.entryPoint.queryFilter(
      this.entryPoint.filters.UserOperationEvent(userOpHash)
    )
```

souradeep-das: yes this would look for the exact event (with the above filter) from the start of the chain until the latest block. Since the response will consist only of one event, some L1 providers eth_getLogs would work because they sometimes only have a limit on the response size. For boba, however this would fail because it wants a block range. 

Hmm true, I was trying to delegate the task of specifying the range to the user, since the bundler does not directly use this, but valid point - the user might not be aware. But otherwise, if we split and do the searching on the bundler, that might be multiple requests for each user request. does that open any DoS angle? maybe more with Promise.all?

packages/boba/bundler/src/modules/EventsManager.ts
```diff
@@ -15,13 +16,18 @@ const debug = Debug('aa.events')
...
+  provider: JsonRpcProvider
   lastBlock = 0
 
   constructor(
     readonly entryPoint: EntryPoint,
     readonly mempoolManager: MempoolManager,
-    readonly reputationManager: ReputationManager
-  ) {}
+    readonly reputationManager: ReputationManager,
+    readonly l1Offset: number
```
souradeep-das: woops I'll change this to l2Offset and make "5000" configurable

packages/boba/bundler/src/BundlerServer.ts
souradeep-das: as per our discussions, created an issue for this here: https://github.com/bobanetwork/boba/issues/865, which I will tackle later. Thanks for pointing this out guys!

packages/boba/bundler/src/modules/EventsManager.ts
InoMurko: https://github.com/bobanetwork/boba/issues/866

InoMurko (approved):

souradeep-das:
converted the log chunksize to a configurable value, please review the last two commits. (associated docs updated here https://github.com/bobanetwork/boba/pull/867)

packages/boba/bundler/src/UserOpMethodHandler.ts
```diff
@@ -227,11 +227,14 @@ export class UserOpMethodHandler {
   }
 
   async _getUserOperationEvent(
-    userOpHash: string
+    userOpHash: string,
+    opts?: any
```
wsdt, line 231: Should we add a type for this?

wsdt (approved):
left a comment, looks good

PR #907, opened 2023-06-07
split PR commenter bc. of changed flow

wsdt:
## Overview

As we seemingly now create a new PR on merge into develop for bundler_sdk, and changed the trigger from pullrequest to onpush, we need to remove the pull request commenter as there is no PR. 

## Changes

- Split up into separate action now. 

## Testing

Test PR

### DIFF
--- a/.github/workflows/aa-bundler-sdk-pr.yml
+++ b/.github/workflows/aa-bundler-sdk-pr.yml
@@ -1,7 +1,7 @@
 name: Will trigger version bump of AA-Bundler for NPM
 
 on:
-  push:
+  pull_request:
     branches:
       - develop
     paths:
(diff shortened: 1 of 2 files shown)


wsdt:
Saw you changed the flow to create a new PR when we merge into develop, just adapted the PR commenter. 
Feel free to close this PR if you think we don't need a PR comment as we create a new PR after merging into develop anyway. 

@InoMurko

InoMurko:
The flow didn't change, it should always behave like that.
The stuff that goes into master comes from develop. We can't have different code in master.

So if SDK needs a bump, it needs to be on top of develop, which then goes into master as a release PR.

wsdt:
> The flow didn't change, it should always behave like that. The stuff that goes into master comes from develop. We can't have different code in master.
> 
> So if SDK needs a bump, it needs to be on top of develop, which then goes into master as a release PR.

sure sure that's fine, I'm referring to the PR comment that we had before that comments on PRs merging into develop to inform you about the release trigger. 

But since the action now listens on push it can't add a PR comment obviously, because there is no PR. That's the issue :-)

InoMurko:
```
As we seemingly now create a new PR on merge into develop for bundler_sdk, and changed the trigger from pullrequest to onpush, we need to remove the pull request commenter as there is no PR.
```
what does thsi comment mean?
- create a new PR on merge into develop for bundler_sdk **yes**
- we need to remove the pull request commenter as there is no PR **what does this mean?**

wsdt:
The action added this PR comment when the PR contained commits that would trigger a bundler_sdk release: 

```
 ⚠️ **This PR triggers a minor release of the Bundler SDK** (Account Abstraction) to NPM as you made changes in the corresponding `bundler_sdk` directory.

              Please make sure this is intentional.

              The package that is going to be updated is `@bobanetwork/bundler_sdk`.
```

But we now create a new PR on push to develop, meaning the action trigger has changed from "on_pullrequest" to "on_push". Thus, making it impossible to add the PR comment above in that very same GH action. 

That's why we have to split the action up.

InoMurko (approved):

PR #939, opened 2023-06-09
Add AA Examples

wsdt:
## Overview

Add 2 AA examples (high level and low level api)

Note: Used the local packages (which runs through successfully) - but replaced with version numbers for remote packages to be a viable example. Because of sdk change etc. the examples only work with the upcoming release once it is merged. Due to `^` version annotation it would automatically use the newest version then. 

## Changes

- Add high level example
- Add low level examples
- Add block range config to bundler sdk (as needed to avoid blockrange exceeded error)

## Testing

Script & test running through successfully on Boba BNB testnet.


InoMurko (approved):
Goodio!

boba_examples/aa-high-level-example/scripts/index.js
```diff
@@ -0,0 +1,145 @@
+"use strict";
```
souradeep-das: tsc generated file?

boba_examples/aa-high-level-example/README.md
```diff
@@ -0,0 +1,11 @@
+# Account Abstraction example
+
+Example for Boba BNB testnet to send userOperation to a wallet through the bundler (high level api).
+This example conducts a simple ETH transfer.
```
souradeep-das: simple contract call?

boba_examples/aa-low-level-example/scripts/index.js
```diff
@@ -0,0 +1,145 @@
+"use strict";
```
souradeep-das: tsc generated file?

boba_examples/aa-low-level-example/scripts/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    to: account,
+  })
+
+  const accountAPI = new SimpleAccountAPI({
+    provider: local_provider,
+    entryPointAddress,
+    owner: l2Wallet,
+    accountAddress: account,
+  })
+
```
souradeep-das, line 65: some comment logs on the intermediary steps could be useful to the user, when they are waiting for the script to run? 
for ex - generating userOp, signing etc

souradeep-das (approved):
🚀

boba_examples/aa-high-level-example/scripts/index.js
wsdt: yes!

PR #981, opened 2023-06-21
Feat/background page title plus

alvaro-ricotta:

## Overview

Background component has been added
PageTitle component has been added

## Changes

We added two new components:

Background component: It handles the gradient and a small animation of lines.
PageTitle component: It displays global titles on each necessary page.
PageTitle was developed using TypeScript (ts), Jest, and Storybook, so it includes testing and documentation capabilities. In the case of Background, it is a simple styled component and doesn't require these additional tools.

We added a **hard reset** in the global style of the theme to avoid design inconsistencies and issues across different browsers.

The Header component was updated to include a fixed background in both themes.


sk-enya:
Review comments:
1. Looks background design is not matching with designs
2. Looks background checks layout seems to be rolling down countinously like an animation!

packages/boba/gateway/src/components/global/background/styles.ts
```diff
@@ -0,0 +1,75 @@
...
+  height: 100%;
+  position: absolute;
+  z-index: 1;
+  top: -50vh;
+  background-image: ${(props) => gradientColor(props.theme.name)};
+  background-clip: padding-box;
+  background-origin: content-box;
+`
+
+const playAnimation = keyframes`
```
sk-enya, line 56: Is it mentioned somewhere to have animation?

packages/boba/gateway/src/components/global/PageTitle/index.tsx
```diff
@@ -1,13 +1,20 @@
...
 
+import { PageTitleTypes } from './types'
 import { PageTitleContainer, Title, Slug } from './styles'
+import { pageTitleWhiteList } from './constants'
 
 export const PageTitle: FC<PageTitleTypes> = ({ title, slug }) => {
+  const location = useLocation()
+  const currentPath = location.pathname
+
+  const { title: pageTitle, slug: pageSlug } =
```
sk-enya: This is fine as far as we have pages related title and slug.

But let's consider in case of bridge page we don't have this so doesn't make sense to render the compoenent with empty data. 

let's return the null or just <></> if no path matches with config.

packages/boba/gateway/src/components/layout/PageTitle/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+  const { title: pageTitle, slug: pageSlug } =
+    pageTitleWhiteList.find((page) => page.path === currentPath) || {}
+  return (
+    <PageTitleContainer>
+      <Title variant="h1">{pageTitle || title}</Title>
+      <Slug variant="body1">{pageSlug || slug}</Slug>
+    </PageTitleContainer>
+  )
+}
```
sk-enya: This is fine as far as we have pages related title and slug.

But let's consider in case of bridge page we don't have this so doesn't make sense to render the compoenent with empty data.

let's return the null or just <></> if no path matches with config.

sk-enya (changes requested):

sk-enya:
@alvaro-ricotta I haven't found any changes for the background for bridge & other pages. 
can you please fix those as well?

packages/boba/gateway/src/components/layout/PageTitle/index.tsx
alvaro-ricotta: MMM, it's a PageTitle right? There should be a path/title available there.

alvaro-ricotta:
> @alvaro-ricotta I haven't found any changes for the background for bridge & other pages. can you please fix those as well?

I just pushed the update for this requeriment

PR #983, opened 2023-06-22
dont run if you're not changing specific code

InoMurko:
## Overview

To preserve CircleCI credits and quicker feedback on tests, integration tests and proxyd tests should only be run if they match changes from the `check-changed` parameter

## Changes

Added check-changed
- we dont want to run integration tests if changes are on the gateway

## Testing

/


wsdt (approved):

.circleci/config.yml
```diff
@@ -53,6 +65,9 @@ commands:
         description: Folder to run tests in
         type: string
     steps:
+      - check-changed:
+          patterns: common-ts,contracts,core-utils,data-transport-layer,message-relayer,regenesis-surgery,sdk,integration-tests,l2geth,packages/boba/account-abstraction,packages/boba/bobalink,packages/boba/bundler,packages/boba/bundler_sdk,packages/boba/bundler_utils,packages/boba/contracts,packages/boba/gas-price-oracle,packages/boba/register,packages/boba/teleportation
```
souradeep-das: should we add ops, ops_boba?

souradeep-das (approved):

.circleci/config.yml
InoMurko: nice find

PR #1007, opened 2023-07-03
Jemeza/gtwy 2.5 history

jemeza:

## Overview

This PR updates the history page for gateway 2.0

## Changes

- Rewrite whole history page.
- Added all networks and corresponding call to load history transactions 
- Disable date picker

## Testing

WIP


packages/boba/gateway/src/components/global/svg/index.tsx
```diff
@@ -2,12 +2,16 @@ import React from 'react'
 import { ReactSVG } from 'react-svg'
 import { SvgTypes } from './types'
 
-export const Svg: React.FC<SvgTypes> = ({ src, fill }) => (
+export const Svg: React.FC<SvgTypes> = ({ src, className, fill, alt }) => (
```
alvaro-ricotta: If you changed the props of a component, make sure to update the changes in Storybook, and if it has/comes with tests, ensure to add the test cases.

packages/boba/gateway/src/components/global/svg/types.ts
```diff
@@ -1,4 +1,6 @@
 export interface SvgTypes {
   src: string
-  fill: string
+  fill?: string
+  className?: string
+  alt?: string
```
alvaro-ricotta: ALT is not a valid attribute in an SVG. If you want, you can use a hidden TITLE for SEO purposes. However, it's important to note that our dapp does not have server-side rendering, so it's not required.

packages/boba/gateway/src/components/global/table/styles.ts
```diff
@@ -0,0 +1,27 @@
...
+import { Row } from 'components/global/containers'
+
+export const TableHeaderContainer = styled(Row)`
+  padding: 20px;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  border-radius: 6px 6px 0 0;
+  background: ${(props) => props.theme.colors.gray[800]};
+  @media (max-width: 960px) {
```
alvaro-ricotta: We are going to need responsive tags throughout the app. Are you up for creating a hook? This way, the implementation would look similar to:

${mobile} {
...code
}

${tablet} {
...code
}

and we avoid defining different queries throughout each styled component.

packages/boba/gateway/src/containers/history/History.js
```diff
@@ -66,52 +74,106 @@ function History() {
...
-        accountEnabled={layer}
-      />
-
       {layer && (
         <>
           <S.Header>
-            <div className={styles.searchInput}>
+            <div
+              className={
+                theme.palette.mode === 'light'
```
alvaro-ricotta: Do not use ternaries in the code, and avoid using inline styles. Please move this code to a styled component.

packages/boba/gateway/src/components/global/table/styles.ts
sk-enya: Yeah, that's a good point that must be needed. @jemeza can you please prepare the same kind of structure and get it added.

packages/boba/gateway/src/util/objectManipulation.ts
```diff
@@ -0,0 +1,3 @@
+export const validArray = (array: any): any[] => {
+  return array.filter((element: any) => element)
+}
```
sk-enya: Are you sure this is the way to validate the wether it's Array or not?

Better to make use of. [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray)

```
Array.isArray(inputArray)
```

packages/boba/gateway/src/selectors/transactionSelector.js
```diff
@@ -10,26 +10,28 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License. */
 
-export function selectTransactions (state) {
-  return Object.values(state.transaction)
+import { validArray } from 'util/objectManipulation'
+
+export function selectTransactions(state) {
+  return validArray(Object.values(state.transaction))
```
sk-enya: It's good to move this validation at the places where we are using this state value instead of adding logics in the selectors itself.
Let's refactored this make selectors are just for selecting the value out of state.

packages/boba/gateway/src/containers/history/History.styles.js
```diff
@@ -1,11 +1,12 @@
 import styled from '@emotion/styled'
```
sk-enya: Please refactored this styles.js with `styles.ts` import the styled from `styled-components` so we can make use of our theme here.

Please rewrite whole file and it's import in history.js

packages/boba/gateway/src/containers/history/History.styles.js
```diff
@@ -1,11 +1,12 @@
...
 import { Box } from '@mui/material'
```
sk-enya: Let's remove use of `Box` from material UI prepare styles with div.

packages/boba/gateway/src/containers/history/History.styles.js
```diff
@@ -82,5 +88,70 @@ export const LayerAlert = styled(Box)(({ theme }) => ({
   [theme.breakpoints.up('md')]: {
     padding: '25px 50px',
   },
+}))
+
+export const Table = styled(Box)(({ theme }) => ({
```
sk-enya: Let's move this to styles.ts and `import styled from styled-components` and please do not make use of Box from material UIl.

packages/boba/gateway/src/containers/history/History.js
```diff
@@ -13,51 +13,59 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 import { fetchTransactions } from 'actions/networkAction'
 
-import Exits from './TX_Exits'
-import Deposits from './TX_Deposits'
-import All from './TX_All'
-import Pending from './TX_Pending'
-import Transfers from './TX_Transfers'
-
 import * as S from './History.styles'
 import styles from './TX_All.module.scss'
```
sk-enya: Let's remove use of module.scss and move all the styles with styled component so we can use the theme correctly .

packages/boba/gateway/src/containers/history/constants.ts
```diff
@@ -0,0 +1,38 @@
+import AllNetworksIcon from '../../images/allNetworks.svg'
+
+export const ALL_NETWORKS = {
```
sk-enya: Please add types / interface to this constants and below others.

packages/boba/gateway/src/components/filter/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+    document.addEventListener('mousedown', handleClickOutside)
+    return () => {
+      // Unbind the event listener on clean up
+      document.removeEventListener('mousedown', handleClickOutside)
+    }
+  }, [dropdownRef])
+
+  useEffect(() => {
+    setSelectedItem(defaultItem)
+  }, [defaultItem])
```
sk-enya: Is `defaultItem` also getting changed from parent? if not I don't think this hook is required.

packages/boba/gateway/src/components/filter/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  return (
+    <DropdownContainer className={`dropdown ${className}`} ref={dropdownRef}>
+      <Header
+        onClick={handleDropdown}
+        error={error}
+        className={
+          isOpen ? `dropdown ${className} active` : `dropdown ${className}`
+        }
```
sk-enya: Let's move this code style.ts itself.

```
export const Header = styled.div.attrs<{ error: boolean, className, isOpen }>({
  className: `dropdown ${className} ${isOpen ? 'active' : ''}`
})``

```

packages/boba/gateway/src/components/filter/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+          <DropdownContent>
+            {items.map((item, index) => (
+              <Option
+                key={index}
+                className={
+                  item.value &&
+                  (item.value === selectedItem.value
+                    ? `dropdown ${className} active`
+                    : `dropdown ${className}`)
+                }
```
sk-enya: You can also move this className down to styles.ts  files with logic mention above so it will be much cleaner.

sk-enya (changes requested):

packages/boba/gateway/src/components/filter/index.tsx
jemeza: that's super cool, will do

packages/boba/gateway/src/util/objectManipulation.ts
jemeza: it wasn't checking wether it was valid array or not but rather extracting valid values

jemeza: i did move the logic

sk-enya (approved):
Looks good!

Issue #996, opened 2023-06-29
Faucet redesign


wsdt:
## Issue Type

```
[ ] bug report
[ x] feature request
```

## Current Behavior
Twitter API faucets don't work without premium anymore. Switched to local file system based verification temporarily. 
Consider proper redesign, captcha, other social media verification etc. 

Related to #995 #989

InoMurko:
FYI: the faucet redesigned will be used on **mainnets** as well. Do they need a new design? @thenursegreg42069

alvaro-ricotta:
FYI: we can use discord instead, only for verified accounts with mobile phone + email

InoMurko:
That sounds too intrusive. I wouldn’t give my phone number

alvaro-ricotta:
You should not request anything, just allow access to the service for individuals with KYC. This way, we prevent alt accounts from draining our faucet.

wsdt:
We could check if Discord member yes/no then x tokens per 24h.

InoMurko:
https://gnosisfaucet.com/ this looks simple and OK to me

thenursegreg42069:
yes we would like to have myself and curtis review the user facing design once it is complete and before any public release

I've seen discord faucets work just fine in the past if you allow only verified accounts

alvaro-ricotta:
@InoMurko https://github.com/Hammad69275/NoCaptchaSolver

wsdt:
if we go for a non HybridCompute approach anyway, we could also consider the PoWFaucet (guess one of the dev-friendliest)

wsdt:
> https://gnosisfaucet.com/ this looks simple and OK to me

starting with HCaptcha as suggested by Ino. 
Please let me know if you prefer an alternative.

wsdt:
closing as completed, opening a Frontend issue

PR #1020, opened 2023-07-11
Simple Captcha faucet

wsdt:
:clipboard: Related to #996 

## Overview

Chain agnostic faucet.

Based on previous discussions and the now paid Twitter API, I created a super simple lambda faucet that uses HCaptcha and has a 24h limit on it. 

Lambda & API Gateway already running on AWS. 

Before we deploy this, we should integrate & test it on the running frontend. 

The lambda expects the following request: 

```json
{
 "hcaptcha_resp": "{CAPTCHA_RESPONSE_TO_VALIDATE}",
 "network": "{BOBA_BNB_TESTNET | BOBA_GOERLI_TESTNET | BOBA_AVAX_TESTNET}", 
 "wallet_addr": "{WALLET_ADDRESS_TO_SEND_FUNDS_TO}"
}
```
Private key of faucet (as always) should only contain testnet funds. 

Current issuance: 
* Alt L2s: Native amount (BOBA) = 1 BOBA, token amount (e.g. tBNB) = 0.1 
* Goerli: Eth = 0.1, Boba token = 1

## Testing
Ran test events on AWS. 

Exemplary tx on boba bnb testnet (currently two tx, one for native, one for token): 
https://blockexplorer.testnet.bnb.boba.network/tx/0xecd44711333f8fa49ddd293084bc4487e97b04e9dd7ffd1fa3f3259d79479dc2


boba_community/hc-captcha-faucet-2023/captcha.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+# or...
+# authorized_contract = '0xOF_YOUR_HELPER_CONTRACT' # to restrict access to only your smart contract
+
+# NOTE: When taking the data payload from the original event for debugging then remove the first 64 bits!
+def lambda_handler(input, context):
+  print("DEBUG: ", input)
+
+  print("DEBUG: from Geth:", input)
```
InoMurko:
```suggestion
  print("DEBUG: from L2:", input)
```

boba_community/hc-captcha-faucet-2023/captcha.py
```diff
@@ -0,0 +1,115 @@
...
+        "txId": txId
+      })
+    }
+  else:
+    returnPayload = {
+      'statusCode': 400,
+      'body': json.dumps({
+        "error": error_msg
+      })
+    }
```
InoMurko: I would refactor this - extract the validation part that either makes or breaks claiming funds once every 24 hours, and then the separate part is response.
- captcha solving
- validation
- response

boba_community/hc-captcha-faucet-2023/captcha.py
```diff
@@ -0,0 +1,115 @@
...
+
+  allowed_to_claim = time_limit_ok and check_hcaptcha(hcaptcha_client_response)
+  if not allowed_to_claim:
+    if not time_limit_ok:
+      error_msg = "Wait 24h"
+    else:
+      error_msg = "Captcha failed"
+
+  if allowed_to_claim:
+    # TBD: potentially add rpc url whitelist
```
InoMurko: wdym?

InoMurko (approved):
left a few nitpicks re style, but good stuff

boba_community/hc-captcha-faucet-2023/captcha.py
wsdt: this faucet is chain agnostic, basically you can send any provider url to the faucet and issue ETH/BOBA there. 
From a security standpoint I just noted that we might want to consider a provider whitelist to ensure that only pre-defined testnets work (and under no way mainnet funds, although why should we send real funds to this PK).

wsdt: added simple whitelist

souradeep-das (approved):
looks great!

PR #1056, opened 2023-07-24
V2 earn stake

alvaro-ricotta:
## Overview

1. Stake logic moved to StakeModal
2. PlaceholderConnect component has been created. 
3. New layout and design for Stake
4. All files has been migrated to ts + styled-components

Screenshots : 

image
image
image


packages/boba/gateway/src/components/layout/Header/__snapshots__/index.test.tsx.snap
```diff
@@ -51,7 +51,7 @@ exports[`Layout => Header should match the snapshot 1`] = `
       class="sc-fABZjn daRNmZ"
     >
       <button
-        class="sc-kuWgmH gPrhcx"
```
sk-enya: Please check once this should not have to be change as far as the change in class names.

packages/boba/gateway/src/components/modal/Modal.tsx
```diff
@@ -27,9 +27,9 @@ import {
 import { Typography } from 'components/global/typography'
 import * as S from './styles'
```
sk-enya, line 21: Does it take too much time to convert these styles to `styles.ts` with `styled-component`?

packages/boba/gateway/src/components/modal/Modal.tsx
```diff
@@ -39,44 +39,39 @@ const _Modal = ({
...
-      ismobile={isMobile ? 1 : 0}
-      slots={{
-        backdrop: S.Backdrop,
-      }}
       disableAutoFocus={true}
     >
       <Fade in={open}>
         <Container sx={{ maxWidth: '450px !important' }}>
-          <S.Style minHeight={minHeight || '450px'} isMobile={isMobile} transparent={!!transparent || !!isMobile}>
+          <S.Style transparent={!!transparent || !!isMobile}>
```
sk-enya, line 51: Seems like you are using the wrong variable name it should say `isMobileDevice`

packages/boba/gateway/src/components/modal/Modal.tsx
```diff
@@ -39,44 +39,39 @@ const _Modal = ({
...
   minHeight,
-  newStyle = false
-}) => {
-  const theme = useTheme();
-  const isMobile = useMediaQuery(theme.breakpoints.down('md'));
+  isMobile = false,
+  newStyle = false,
+}: ModalInterface) => {
+  const theme = useTheme()
+  const isMobileDevice = useMediaQuery(theme.breakpoints.down('md'))
```
sk-enya, line 39: It's better to handle it via `styles.ts` instead of loading it from material

packages/boba/gateway/src/components/modal/types.ts
```diff
@@ -0,0 +1,12 @@
+export interface ModalInterface {
+  children: any
+  isMobile?: boolean
+  open: any
+  onClose: any
```
sk-enya, line 5: Should say function instead of any.

packages/boba/gateway/src/components/stake/transactionList/index.tsx
sk-enya, line 1: This folder name is supposed to say `StakeList` instead of `transactionList` as it's more about stake not history page txs

packages/boba/gateway/src/components/stake/transactionList/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { Button } from 'components/global/button'
+import { Typography } from 'components/global/typography'
+import { ModalTypography } from 'components/global/modalTypography'
+
+import { TransactionListInterface } from './types'
+import { StakeItemDetails, Token } from './styles'
+import { getCoinImage } from 'util/coinImage'
+
+const TransactionList = ({ stakeInfo }: TransactionListInterface) => {
+  const [state, setState] = useState({
```
sk-enya: Can you please give better name instead of state as it might be confusing longer run with redux/state.

packages/boba/gateway/src/components/stake/transactionList/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+      </div>
+    </StakeItemDetails>
+  )
+}
+
+const mapStateToProps = (state: any) => ({
+  fixed: state.fixed,
+})
+
+export default connect(mapStateToProps)(TransactionList)
```
sk-enya: Why do we have connect here?
We can directly read the state from the redux store using `useSelector`. and even it looks like you missed remove this code as `props.fixed` is not getting used anywhere in the file.

packages/boba/gateway/src/components/stake/transactionList/types.ts
```diff
@@ -0,0 +1,10 @@
+type TransactionType = {
```
sk-enya, line 1: should rename to `StakeType` & `IStakeTypeList`.

packages/boba/gateway/src/containers/modals/stake/DepositStake.tsx
```diff
@@ -0,0 +1,270 @@
...
+}
+
+const mapStateToProps = (state: any) => ({
+  ui: state.ui,
+  fixed: state.fixed,
+  setup: state.setup,
+  balance: state.balance,
+})
+
+export default connect(mapStateToProps)(DepositStake)
```
sk-enya: Why do we need to use this?
Can't we use directly the useSelector to get the value of all states from redux? as it's functional component.

sk-enya (changes requested):

sk-enya:
In design snapshot share above it seems `Stake Button` needs to be full width.

packages/boba/gateway/src/components/modal/Modal.tsx
alvaro-ricotta: Don´t worry about it i´ll refactory all entire mobile-/table styles to styledcomponents

packages/boba/gateway/src/components/stake/transactionList/index.tsx
alvaro-ricotta: For that same reason, there is the Stake folder; if I'm referring to TransactionList inside Stake, I'm talking about the StakeTransactionList, but if you prefer, I can rename it.

packages/boba/gateway/src/components/stake/transactionList/index.tsx
alvaro-ricotta: Done

packages/boba/gateway/src/components/stake/transactionList/index.tsx
alvaro-ricotta: Done

packages/boba/gateway/src/containers/modals/stake/DepositStake.tsx
alvaro-ricotta: Done

packages/boba/gateway/src/components/stake/transactionList/index.tsx
sk-enya: yeah, can you please rename it? will help us to cleanup the code base and reduce confusion in future.

packages/boba/gateway/src/components/modal/Modal.tsx
sk-enya: It's more about the variable name not about the refactoring, please read the variable name `isMobile` which is not defined above.

sk-enya (approved):

PR #1083, opened 2023-08-05
Gateway: Teleportation integration

### LINKED ISSUE #1034: Integrate Teleportation in Gateway 2.5
Teleportation contract: 
https://github.com/bobanetwork/boba/blob/develop/packages/boba/contracts/contracts/Teleportation.sol

Current deployments on testnets: 
https://github.com/bobanetwork/boba/tree/develop/packages/boba/teleportation

Deployed contracts have not yet any routes configured, what routes do we want to provide liquidity for @InoMurko (on testnet)?

Not deployed on Mainnet yet. 

---

User flow: 

1. User calls this functions via the Gateway. 
```sol
 /**
     * @dev Accepts deposits that will be disbursed to the sender's address on target L2.
     * The method reverts if the amount is less than the current
     * minDepositAmount, the amount is greater than the current
     * maxDepositAmount.
     *
     * @param _token ERC20 address of the token to deposit.
     * @param _amount The amount of token or native asset to deposit (must be the same as msg.value if native asset)
     * @param _toChainId The destination chain ID.
     */
 function teleportAsset(address _token, uint256 _amount, uint32 _toChainId)
```

Note: Only whitelisted routes are allowed, otherwise the transaction will revert. Meaning: if the destination chain & token are not both supported from the source chain, it cannot be bridged. Which assets are whitelisted has to be maintained in the frontend for now or by reading events from the contract: 

```sol
event TokenSupported(
        address indexed token,
        uint32 indexed toChainId,
        bool supported // if true, then this route is supported (from contract on source chain to destination (to) chain)
    );
```

Within a short period of time the Teleportation service on AWS should disburse the asset on the destination chain via the deployed Teleportation contract by calling: `function disburseAsset(Disbursement[] calldata _disbursements)`. 

After this event on the destination chain has been fired, you can notify the user on the gateway, that the funds have arrived: 

```sol
  event DisbursementSuccess(
        uint256 indexed depositId,
        address indexed to,
        address indexed token,
        uint256 amount,
        uint32 sourceChainId
    );
```
----

Error handling for native assets, in that case we could/should even do a retry mechanism in the gateway. 

```sol
  event DisbursementSuccess(
        uint256 indexed depositId,
        address indexed to,
        address indexed token,
        uint256 amount,
        uint32 sourceChainId
    );

    /** @dev only for native assets */
    event DisbursementFailed(
        uint256 indexed depositId,
        address indexed to,
        uint256 amount,
        uint32 sourceChainId
    );

    /** @dev Only for native assets */
    event DisbursementRetrySuccess(
        uint256 indexed depositId,
        address indexed to,
        uint256 amount,
        uint32 sourceChainId
    );
```

wsdt:
Closes #1034 

# What is being done?
1. Added Teleportation Support on Gateway (as discussed 3rd tab for Teleportation only shown on testnet)
2. History page for Teleportation implemented
3. Implemented Teleportation for GraphQL, deployed for testnets (can be teared down once in production to save infra costs)

**PLEASE NOTE**: This PR is dependant (and therefore based on #1090 )

Right now GraphQL endpoints have been deployed on AWS without specific DNS - as long as we don't tear down the instances we should keep the AWS native DNS (e.g. loadB-sldkfj, etc.). 

Will catch up with DevOps, to get some testnet DNS subdomains for this.

Mainnet (Teleportation disabled): 
image

Testnet: 
image


package-lock.json
boyuan-chen, line 1: Can we remove this file? Since we use `yarn` to the package management.

packages/boba/contracts/teleportation_deploy_logs/.gitkeep
boyuan-chen, line 1: Can we delete this empty file?

packages/boba/subgraph/L2/config/local.json
boyuan-chen, line 1: What's this used for?

packages/boba/teleportation/src/utils/chains.ts
```diff
@@ -81,12 +81,25 @@ export const BobaChains: IBobaChains = {
...
+    supportedAssets: {
+      /*'0x0000000000000000000000000000000000000000': 'ETH',
+      '0x42bBFa2e77757C645eeaAd1655E0911a7553Efbc': 'BOBA',
+      '0xdAC17F958D2ee523a2206206994597C13D831ec7': 'USDT',
+      '0xB8c77482e45F1F44dE1745F52C74426C631bDD52': 'BNB',*/
+    },
+  },
   5: {
-    url: 'https://goerli.gateway.tenderly.co',
+    url: 'https://ethereum-goerli.publicnode.com', //'https://goerli.gateway.tenderly.co',
```
boyuan-chen: Why don't use our tenderly RPC endpoint?

boyuan-chen (commented):
It seems that the gateway doesn't work. 
```
NS: ERROR: InitializeBase  Error: invalid contract address or ENS name (argument="addressOrName", value=undefined, code=INVALID_ARGUMENT, version=contracts/5.7.0)
```

boyuan-chen:
The integration tests need to be fixed.

souradeep-das:
not sure if this is the right place, can we somehow add a small alert/msg to let users know - that they should not use an unconventional wallet/multisig (if they are using a contract account as recipient) with a costly fallback method. (since we pass 3000 gas for disbursements

packages/boba/subgraph/L2/config/local.json
wsdt: For testnet deployments & a local graph-node to debug on networks that are not supported.

packages/boba/contracts/teleportation_deploy_logs/.gitkeep
wsdt: I'm logging the Teleportation routes there (to have a file that logs them). Especially for mainnet deployment then. Otherwise this would need to be read on-chain.

wsdt:
> It seems that the gateway doesn't work.
> 
> ```
> NS: ERROR: InitializeBase  Error: invalid contract address or ENS name (argument="addressOrName", value=undefined, code=INVALID_ARGUMENT, version=contracts/5.7.0)
> ```

If I check out the current branch it works. I guess this is because the @boba/register package has not been rebuilt? 

In the PR I had to change the addresses from `Proxy__Teleportation` to `Proxy_L2Teleportation` etc. 

"Proxy__L1Teleportation": "0x7f6a32bCaA70c65E08F2f221737612F6fC18347A",
  "Proxy__L2Teleportation": "0xf4d179d3a083Fa3Eede935FaF4C679D32d514186",
  "L1Teleportation": "0xD151c8F0dc69618e6180a2dC74B05cCE3E08e0aC",
  "L2Teleportation": "0x46FA6144C61d2bb9aCDc3Ca90C8673dd9B6caEB2"

wsdt:
> The integration tests need to be fixed.

fixed thx!

wsdt:
> It seems that the gateway doesn't work.
> 
> ```
> NS: ERROR: InitializeBase  Error: invalid contract address or ENS name (argument="addressOrName", value=undefined, code=INVALID_ARGUMENT, version=contracts/5.7.0)
> ```

was a mainnet issue, since Teleportation is not deployed there yet. Fixed good catch!

wsdt:
@souradeep-das 
image

boyuan-chen:
I found another bug of gateway. It didn't work on my end and I got: 
```
NS: ERROR: InitializeBase  TypeError: Cannot read properties of undefined (reading 'payloadForL1SecurityFee')
```
Is this bug from the develop branch?

wsdt:
> payloadForL1SecurityFee

Just checked, I don't get that issue. 

And the code line that probably causes this is: 

**NetworkService.js:**
```js

 const networkDetail = getNetworkDetail({
      network,
      networkType
    })

if (NETWORK[network]) {
        this.payloadForL1SecurityFee = networkDetail.payloadForL1SecurityFee

```

Because networkDetail might be undefined for you for whatever reason? Just my guess. 

The `getNetworkDetail` function has not changed since March 2023 as well. So yes the code that may cause this, should also be in develop if I didn't overlook something.

Is the issue deterministic and still there? 

Thank you!

boyuan-chen:
> > payloadForL1SecurityFee
> 
> Just checked, I don't get that issue.
> 
> And the code line that probably causes this is:
> 
> **NetworkService.js:**
> 
> ```js
>  const networkDetail = getNetworkDetail({
>       network,
>       networkType
>     })
> 
> if (NETWORK[network]) {
>         this.payloadForL1SecurityFee = networkDetail.payloadForL1SecurityFee
> ```
> 
> Because networkDetail might be undefined for you for whatever reason? Just my guess.
> 
> The `getNetworkDetail` function has not changed since March 2023 as well. So yes the code that may cause this, should also be in develop if I didn't overlook something.
> 
> Is the issue deterministic and still there?
> 
> Thank you!

Ok! That's fine. Thank you.

boyuan-chen (approved):

alvaro-ricotta:
hey @wsdt quick question for you : 

image

I'm getting 2 different alerts: 
1) This bridge doesn't support smart-contract wallets that use a costly fallback method.
2) For this asset you need to bridge at least 0.001.

In the case of the second alert, it shouldn't be displayed by default, only when we input a value into the input.

On the other hand, I don't understand what the first alert refers to; I'm using a standard wallet.

image

After clicking on the bridge, which technically takes less than 1 minute (according to the time estimate), it asks me to wait for 64 blocks to confirm. Is this correct? Or should we update the text in case the bridge comes from now? The idea of using teleport is, I assume, to avoid waiting for 64 network confirmations, right?

In a general sense, the experience is incredible. The only detail I want to add is that I tried to bridge USDC on Goerli, and it tells me that the token is not currently supported.


The last detail is that it's not bidirectional. If we don't support assets bidirectionally, we shouldn't display the "now" option for certain assets/networks, right? For example, I can bridge from Goerli to Boba Goerli, but I can't bridge from Boba Goerli to Goerli.

alvaro-ricotta:
@jemeza after using now bridge history has been crashed 
image

alvaro-ricotta:
@wsdt keep in mind that your branch is out-of-date too =)

alvaro-ricotta (approved):

packages/boba/gateway/src/services/networkService.js
```diff
@@ -4744,6 +4669,11 @@ class NetworkService {
     return []
   }
 
+  async getCurrentChainId() {
+    const {chainId} = await this.provider?.getNetwork()
+    return chainId;
+  }
+
```
sk-enya: Good to remove this if we are not using.

packages/boba/gateway/src/hooks/useBridge.ts
```diff
@@ -57,38 +77,36 @@ export const useBridge = () => {
...
-      if (approvalReciept === false) {
+      if (approvalReceipt === false) {
         dispatch(
           openError('Failed to approve amount or user rejected signature')
         )
         return
       }
     }
-    return dispatch(depositL1LP(token.address, amountWei))
+    return dispatch(
```
sk-enya, line 127: @wsdt  Correct me if I am wrong, we are going to add teleportation only for TESTNET?

It seems we have completely removed support for FAST LP bridging for MAINNET, is that correct or we still have to support the LP fast bridging on mainnet?

sk-enya:
@wsdt Is teleportation depends on bobaFeeContract? which is a relayer fee I guess we are not going to use right?

if not, remove this fee from the fee section at the bottom of the bridging container. `xChain Relay Fee - 10.0 BOBA` Also boba balance in MM and Wallet should be the same.
Screenshot 2023-08-31 at 7 42 24 PM

sk-enya:
Values on fee sections needs to be added even selected teleportation it's showing the values of `L1LpContract / L2LpContract`
Screenshot 2023-08-31 at 7 46 49 PM

sk-enya:
Looks like fast bridging is not working, failing with error.
Screenshot 2023-08-31 at 7 39 59 PM
Screenshot 2023-08-31 at 7 40 08 PM

packages/boba/gateway/src/hooks/useBridge.ts
```diff
@@ -97,13 +115,13 @@ export const useBridge = () => {
...
-        receipt = await triggerFastDeposit(amountWei)
+        receipt = await triggerTeleportAsset(amountWei, destChainIdBridge!)
       }
     } else {
       if (bridgeType === BRIDGE_TYPE.CLASSIC) {
         receipt = await triggerExit(amountWei)
       } else {
-        receipt = await triggerFastExit(amountWei)
+        receipt = await triggerTeleportAsset(amountWei, destChainIdBridge!)
       }
```
sk-enya: Better to support the fast bridging on Mainnet we should just extend this function to use the triggerTeleportAsset. like
```
if(BRIDGE_TYPE.TELEPORTATION) {
   receipt = await triggerTeleportAsset(amountWei, destChainIdBridge!)
}
```

sk-enya (changes requested):

sk-enya:
@wsdt I tried teleport bridge what I notice 

1. I am unable see transactions on history page as success dialog says. here is link to txs. https://testnet.bobascan.com/tx/0x6883b2edb41fbce96fc8eaad675eccc3befb133b853791dfb9bc6f22ae4dffd9
2. Still waiting to receive funds on my eth L1 goerli form boba L2  https://goerli.etherscan.io/address/0x81654daa2e297a140ca01f12367a4d9c9fc2bf51
3. As @alvaro-ricotta mentioned above dialog teleportation uses the same fast bridge dialog.

sk-enya:
Overall code looks good :) thanks!

packages/boba/gateway/src/hooks/useBridge.ts
wsdt: yes as discussed, teleportation is only available on testnet in the first release.

wsdt:
@alvaro-ricotta 
Thanks for your feedback: 

1. the info message has been added as per prior feedback, please refer above's messages. There is no really reliable way of showing this info as we don't know if the user has a costly fallback method (if they have a smart wallet) on the destination chain. All we could do is checking if the destination address is a contract or not. 

2. Other UI remarks have been incorporated, thanks!

3. Teleportation is bi-directional, and yes USDC is not supported for now, that's why we show this message. Only BOBA and mostly ETH are supported at the moment.

wsdt:
> @wsdt I tried teleport bridge what I notice
> 
> 1. I am unable see transactions on history page as success dialog says. here is link to txs. https://testnet.bobascan.com/tx/0x6883b2edb41fbce96fc8eaad675eccc3befb133b853791dfb9bc6f22ae4dffd9
> 2. Still waiting to receive funds on my eth L1 goerli form boba L2  https://goerli.etherscan.io/address/0x81654daa2e297a140ca01f12367a4d9c9fc2bf51
> 3. As @alvaro-ricotta mentioned above dialog teleportation uses the same fast bridge dialog.

image

AWS service was in a restart loop, working now. 
The BSC testnet node made the service crash as we hit their free quota, replaced now + added more resilient error handling & redeployment.

sk-enya (approved):
@wsdt  all looks good!!

sk-enya:
looks failing with solidity coverage plugin @wsdt ,
https://github.com/bobanetwork/boba/actions/runs/6086249174/job/16512166025?pr=1083#step:9:1042

wsdt:
> looks failing with solidity coverage plugin @wsdt , https://github.com/bobanetwork/boba/actions/runs/6086249174/job/16512166025?pr=1083#step:9:1042

no was just a timeout from Account Abstraction

PR #1088, opened 2023-08-08
Fix/date range picker history

jemeza:
## Overview
Made new date picker component using https://react-day-picker.js.org/ and added it to the history page with appropriate styles.

## Changes

Made a new component that shows the date and when clicked opens up a calendar that allows the user to choose the date. It can also be used to pick date ranges.

## Testing

Added jest test to ensure component functionality. Test can be run from gateway directory with `yarn test src/containers/history/`


sk-enya (approved):
Looks good!

PR #1096, opened 2023-08-10
Fix/add avalanche transaction history

jemeza:
## Overview

Adds Avalanche transactions to the history page and makes history page more robust.

## Changes

Now checking transactions on all supported networks.
Checks for existence of tokens before using them.
Removes link on transaction hashes that are pending or have failed. 

## Testing

Please run locally.


sk-enya (approved):
Looks good

PR #1099, opened 2023-08-10
Trigger Selectors when account is enabled.

alvaro-ricotta:
## Overview

Update wrong network message update when user is already in same network.


sk-enya (approved):
LGTM

PR #1180, opened 2023-09-05
feat: Airdrop gas via Light Bridge

wsdt:
:clipboard: Related to bobanetwork/gateway#69 

Needs to be based on bobanetwork/boba#1083 

## Overview

Airdrop gas when using Teleportation if certain conditions are met. 

Conditions are: 
- Is Airdrop enabled (e.g. I guess we only want to enable it on L2 and not when bridging back to L1)
- Did user airdrop already within the last x seconds (by default `86400`, 1 day)
- Does user bridge asset that is **not** a native asset on the destination network
- Does user **not** have any native balance on the destination network

## Changes

- Changes in asset mapping
- Added integration tests
- Added airdrop feature to Teleportation service
- No smart contract changes to avoid new audit

## Testing
- Integration tests succeed


packages/boba/teleportation/src/service.ts
```diff
@@ -353,12 +371,114 @@ export class TeleportationService extends BaseService<TeleportationOptions> {
...
+        const lastAirdrop = await lastAirdropRepository.findOneBy({walletAddr: disbursement.addr})
+        const unixTimestamp = Math.floor(Date.now() / 1000)
+
+        const airdropCooldownSeconds = this.options.airdropConfig?.airdropCooldownSeconds ?? '86400'
+        if (!lastAirdrop || BigNumber.from(airdropCooldownSeconds).lt(unixTimestamp - lastAirdrop.blockTimestamp)) {
+
+          let nativeAmount = this.options.airdropConfig?.airdropAmountWei
+          if (!nativeAmount) {
+            // default
+            nativeAmount = ethers.utils.parseEther('0.001')
```
souradeep-das: i think overall we can set the defaults a bit lower for eth l2

packages/boba/teleportation/src/service.ts
```diff
@@ -353,12 +371,114 @@ export class TeleportationService extends BaseService<TeleportationOptions> {
...
+      }
     } catch (e) {
       this.logger.error(e)
     }
   }
 
+  /** @dev Checks if major airdrop eligibility criteria has been met such as not bridging native, has no gas on destination network, bridges enough value, .. */
+  async _fulfillsAirdropConditions(disbursement: Disbursement, assetUsdValues: ITokenUSDPrice|undefined) {
+    const nativeBalance = await this.state.Teleportation.provider.getBalance(disbursement.addr)
+    if (nativeBalance.gt('0')) {
```
souradeep-das: not sure if this is important, just wanted to mention 
how about checking for balance gt (default_amount)

integration-tests/test/eth-l2/teleportation.spec.ts
```diff
@@ -928,4 +940,680 @@ describe('teleportation', () => {
...
+        chainId,
+        blockNumber
+      )
+
+      const postNativeBalance = await ethers.provider.getBalance(address1)
+      const postSignerNativeBalance = await ethers.provider.getBalance(randAddress)
+
+      const gasDelta = ethers.utils.parseEther('0.003')
+      expect(preNativeBalance.sub(postNativeBalance)).to.be.closeTo(amountToBridge, gasDelta)
+      expect(postSignerNativeBalance.sub(preSignerNativeBalance)).to.be.closeTo(
```
souradeep-das: for the recipient part is the gasDelta required? (to make certain no airdrop was received)

integration-tests/test/eth-l2/teleportation.spec.ts
```diff
@@ -928,4 +940,680 @@ describe('teleportation', () => {
...
+      expect(preTokenBalance.sub(postTokenBalance)).to.be.eq(
+        utils.parseEther('20')
+      )
+      expect(postSignerTokenBalance.sub(preSignerTokenBalance)).to.be.eq(
+        utils.parseEther('20')
+      )
+      const gasDelta = ethers.utils.parseEther('0.003')
+      expect(preNativeBalance.sub(postNativeBalance)).to.be.closeTo(airdropConfig.airdropAmountWei, gasDelta)
+      expect(postSignerNativeBalance.sub(preSignerNativeBalance)).to.be.closeTo(
+        airdropConfig.airdropAmountWei, gasDelta
```
souradeep-das: the test case says no airdrop, but are we checking for airdrop here?

souradeep-das (commented):
looks great to me!

integration-tests/test/eth-l2/teleportation.spec.ts
wsdt: yes that's correct, because the first airdrop should succeed and the 2nd should be skipped due to the cooldown period :-)

packages/boba/teleportation/src/service.ts
wsdt: you mean to loosen the condition a bit, so that the user always has at least the airdrop amount? :-) 
thx!

integration-tests/test/eth-l2/teleportation.spec.ts
souradeep-das: oh yes i see thanks!

packages/boba/teleportation/src/service.ts
souradeep-das: yes!

souradeep-das (approved):

packages/boba/teleportation/src/exec/run.ts
```diff
@@ -60,6 +62,22 @@ const main = async () => {
     'teleportation-aws-kms-endpoint',
     env.TELEPORTATION_AWS_KMS_ENDPOINT
   )
+  const TELEPORTATION_AIRDROP_GAS_AMOUNT_WEI = config.str(
+    'teleportation-airdrop-gas-amount-wei',
+    env.TELEPORTATION_AIRDROP_GAS_AMOUNT_WEI || '1000000000000000' // 0.001 eth
```
souradeep-das: another place to update the defaults :)

packages/boba/teleportation/src/service.ts
```diff
@@ -353,12 +371,115 @@ export class TeleportationService extends BaseService<TeleportationOptions> {
...
+      "outputs": [
+        {
+          "name": "",
+          "type": "uint8"
+        }
+      ],
+      "payable": false,
+      "stateMutability": "view",
+      "type": "function"
+    }], this.state.Teleportation.provider)
```
InoMurko: maybe the abi should be pulled out?

InoMurko:
Using just one API is dangeours because of possible skewed results:
`- Does user bridge at least x$ (using Coingecko API, fails silently and ignores check if API call fails). Default: $ 15`
I suggest you remove the coingeko or add another API.

packages/boba/teleportation/src/utils/coingecko.ts
```diff
@@ -0,0 +1,40 @@
...
+    console.error(err)
+  }
+}
+
+export const getUsdValueOfTokenViaIDs = async (tokenIds: string[]): Promise<ITokenUSDPrice|undefined> => {
+  try {
+    const response = await axios.get(`https://api.coingecko.com/api/v3/simple/price?ids=${tokenIds.join(',')}&vs_currencies=usd`)
+    return response.data as ITokenUSDPrice;
+  } catch (err) {
+    // Fail silently, non mandatory check
```
InoMurko: if it's not mandayory, why use it? might be better to remove this outside oracle if it can't be reliable (two or more sources so that possible outliers can be removed)

InoMurko:
```
Does user bridge at least x$ (using Coingecko API, fails silently and ignores check if API call fails). Default: $ 15

```
should be removed from PR description

InoMurko (approved):

souradeep-das (approved):
changes look good to me!

since we know what these erc20 tokens are , we could have a configured min bridge value (that's probably too much trouble if token prices fluctuate), otherwise its okay too, we just gotta make sure "bridge gas is sufficiently > the airdrop amount"
(even for a batched call to bridge scenario)

PR #1196, opened 2023-09-06
Making getBalance in network services more robust

jemeza:
## Overview

Small change that prevents getBalance method in network services from ruining user experience in the case of an error.

## Changes

replacing `Promise.all` with `Promise.allSettled` and logging and filtering errors

## Testing

please run locally


sk-enya (approved):
Looks good!

PR #1198, opened 2023-09-08
End to end testing

### LINKED ISSUE #633: [Gateway] Rework .scss files to styled components
## Issue Type

```
[ ] bug report
[ ] feature request
[x] change request
```

## Current Behavior
While redesigning the Gateway pages I sometimes came across .scss files. Since the project works mainly with styled components, this is often very inconvenient because, for example, you can't use variables from the theme in the scss file and some things are not as easy as working with styled components. It would also improve readability and consistency of the whole codebase.

There would also be a problem with the lightmode/darkmode switcher, because some colors should be changed with theme variables and because you can't use them in the .scss files, this would be the only solution to change the colors, but the @media dark setting only works with the browser settings, so it won't work with our "theme switcher"

For example: For the tooltip component we need different title colors for darkmode and lightmode, in this case there is only one .scss file. I added a "@media (prefers-color-scheme: dark)" for reference, but that would only have an effect if the browser setting is the same as the theme switcher. So if you use the browser setting "darkmode" but the theme is set to "lightmode", a wrong color will be displayed.

```
.title {
  font-size: 14px;
  padding: 13px 16px 12px;
  border-radius: 12px;
  color: rgba(3, 19, 19, 0.85);
}

@media (prefers-color-scheme: dark) {
  .title {
    font-size: 14px;
    padding: 13px 16px 12px;
    border-radius: 12px;
    color: rgba(255, 255, 255, 0.85);
  }
}
```

## Suggested Fix
Rework certain .scss files to styled components

### LINKED ISSUE #651: [Gateway]: Add E2E testing
### Integration / E2E Testing

Make sure the gateway has an E2E test in place, such that the test should run a browser instance by opening the gateway.

### LINKED ISSUE #1207: Gateway: Git Actions to run the Integration test!

jemeza:
## Overview
closes: https://github.com/bobanetwork/boba/issues/633
End to End Testing for page header/footer and the bridge and history pages

## Changes

Added end to end testing for both Bridge and History pages using Cypress with the Synpress plugin.

## Testing

Populate the  local `.env` file as specified in readme. Run `yarn start` in `boba/packages/boba/gateway`, open a different terminal and run `yarn test-e2e` in `boba/packages/boba/gateway`.


InoMurko:
can these tests be executed within github actions please?

sk-enya:
Setting to draft as have to move this code to gateway repository.

sk-enya:
Closing as we have corresponding PR on gateway repo
https://github.com/bobanetwork/gateway/pull/33

PR #1227, opened 2023-09-24
Captcha Mainnet & Testnet faucet

### LINKED ISSUE #1024: Mainnet "Faucet"
## Issue Type

```
[ ] bug report
[x] feature request
```
This feature is planned for Gateway 2.6, with an estimated go-live of late September to early October.

## Current Behavior
Currently there is no easy way for first time Boba users to get gas tokens on the Boba L2s. Our bridge and various 3rd party bridges allow the deposit of multiple non-gas assets such as USDC and others. Additionally, ETH is the default token on Boba ETH, so if users deposit Boba they will not have gas to confirm a fee token switch transaction.

The whole idea of being able to bridge funds but getting stuck without gas on the destination side is a major UX hurdle both in theory and from our real world user reports.

## Expected Behavior
The user should have an easy way to get gas token integrated into the user flow.

There are two primary methods to accomplish this:
1) Faucet
2) Integrated into bridge

After discussion with Kevin, option 2 is strongly preferred as it presents several advantages over option 1
1) Better UX - integrated into bridge directly at the point of user contact. Does not require the user to visit other sites. Is simply a feature addition vs. a new product creation.
2) More economically viable - a true faucet results in us giving money away, albeit small - it is financially net negative to us. Bridge integration deducts a fee from the bridged funds, resulting in net neutral for us.
3) Less exploitable - faucets done through twitter/discord/other can be exploited and drained. Giving gas through the bridge does not allow sybil attacks as it is net neutral.

The cons of this approach are that there is more smart contract and front-end work required, and it will need to be audited.

## Suggested Fix

For examples of the desired design, see the following:
https://theaptosbridge.com/bridge
https://www.synapseprotocol.com/

User Flow:
1) When bridging on the Gateway, user will be given the option of "Receiving Gas on Destination"
2) User can either decline, select from default recommended amounts, or set a custom amount (this should have a maximum value).
3) The value of amount the user elects to receive will be deducted from the transaction as a fee. (we need to consider where we receive these fees and do we convert them back to ETH/BOBA to resupply the bridge or hold them as the original token)
4) The user will see the fee deduction alongside the display of any other fees before signing.
5) The user completes the rest of their bridge transaction as applicable and signs. 
6) The user receives their bridged funds plus their selected gas on the destination chain.

Final Thoughts:
@wsdt will be the main engineering POC on this. @thenursegreg42069 will be the main business contact.

wsdt:
:clipboard: Resolves bobanetwork/gateway#69 
Teleportation mainnet faucet: bobanetwork/boba#1180 

## Overview

Faucet based on captcha

## Changes

- Mainnet & Testnet faucet protected by ImageCaptcha
- Removed previous captcha faucet versions
- Added simple meta tx support
- Added voluntary token issuance support (for testnet, for mainnet we can set this simply to 0 or if no token balance available the faucet will only issue native funds)

## Testing

- Unit tests
- Integration tests

----

@thenursegreg42069 
@sk-enya FYI


l2geth/cmd/faucet/faucet.go
```diff
@@ -384,7 +384,7 @@ func (f *faucet) apiHandler(w http.ResponseWriter, r *http.Request) {
 		if err = conn.ReadJSON(&msg); err != nil {
 			return
 		}
-		if !*noauthFlag && !strings.HasPrefix(msg.URL, "https://gist.github.com/") && !strings.HasPrefix(msg.URL, "https://twitter.com/") &&
+		if !*noauthFlag && !strings.HasPrefix(msg.URL, "https://gist.github.com/") && !strings.HasPrefix(msg.URL, "https://mainnetFaucet.com/") &&
```
InoMurko: are these changes needed?

boba_community/hc-captcha-mainnetfaucet-2023/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@boba/mainnetfaucet",
```
souradeep-das: i think we are changing the package names to 'at' bobanetwork?

.gitmodules
```diff
@@ -2,3 +2,6 @@
 	path = l2geth/tests/testdata
 	url = https://github.com/ethereum/tests
 
+[submodule "boba_community/hc-captcha-mainnetfaucet-2023/lib/forge-std"]
```
souradeep-das: nice! first foundry project in the stack maybe?

boba_community/hc-captcha-mainnetfaucet-2023/contracts/BobaMainnetFaucet.sol
```diff
@@ -0,0 +1,125 @@
...
+import '@openzeppelin/contracts/access/Ownable.sol';
+import './TuringHelper.sol';
+
+contract BobaMainnetFaucet is Ownable {
+    address public NativeAddress = 0x4200000000000000000000000000000000000006;
+    uint256 public nativeFaucetAmount;
+    uint256 public waitingPeriod;
+
+    // Turing
+    address public hcHelperAddress;
```
souradeep-das: minor: do we need this var?

boba_community/hc-captcha-mainnetfaucet-2023/contracts/BobaMainnetFaucet.sol
```diff
@@ -0,0 +1,125 @@
...
+        require(nativeClaimRecords[msg.sender] + waitingPeriod < block.timestamp, 'Invalid request');
+
+        bytes32 hashedKey = keccak256(abi.encodePacked(_key));
+        uint256 result = _verifyKey(_uuid, hashedKey);
+
+        require(result == 1, 'Captcha wrong');
+
+        nativeClaimRecords[msg.sender] = block.timestamp;
+
+        (bool sent,) = (msg.sender).call{gas: SAFE_GAS_STIPEND, value: nativeFaucetAmount}("");
```
souradeep-das: I have a question on user's claiming from the faucet directly

Is the nativeFaucetAmount supposed to be > tx gas cost?
If that's the case, there is the danger of a front-running bot which front-runs other accounts taking input captcha params and uses new accounts everytime.

.gitmodules
wsdt: yes! It's a Hardhat<>Foundry project, to keep full compatibility with the rest of the stack.

boba_community/hc-captcha-mainnetfaucet-2023/contracts/BobaMainnetFaucet.sol
wsdt: yes would be supposed to be larger. 
Is that not what the captcha would be for? Or do you have a suggestion how to prevent this?

souradeep-das: I see, in that case the captcha probably becomes less of a protection because `uuid` and `key` are sent as params. so a front-running bot could just wait for others to solve it and steal these params. I was thinking is there a way we could assign msg.sender for a specific uuid? such that verification is `_verifyKey(uuid, key, msg.sender)`

wsdt: gotcha, good call! will do thanks

boba_community/hc-captcha-metafaucet/api/hc_sendMetaTx.js
wsdt, line 1: Until now we always had Python based lambda functions, I tried to retain that. 

But the `web3py` library doesn't support estimating gas for contract function calls associated to a wallet / signed transactions, for that reason I needed to use the JavaScript ethers library.

boba_community/hc-captcha-metafaucet/api/hc_sendMetaTx.js
```diff
@@ -0,0 +1,132 @@
...
+    return false;
+  }
+  const recoveredAddr = ethers.utils.verifyMessage(nonce, sig);
+  return recoveredAddr === address;
+}
+
+async function issue_nonce(redisClient, address) {
+  let nonceToSign = ethers.utils.base58.encode(ethers.utils.randomBytes(16));
+  if (nonceToSign.length !== 22) {
+    nonceToSign = ethers.utils.base58.encode(ethers.utils.randomBytes(20)).substring(0, 22)
```
wsdt, line 127: not necessarily important, but there is a tiny flake (sometimes the nonce not being exactly at 22 string length. 
Although this shouldn't matter at all, I prefer to have it consistent.

boba_community/hc-captcha-metafaucet/api/hc_sendMetaTx.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+  await redisClient.set(address, nonceToSign, 'EX', 600);
```
wsdt, line 129: nonce expires automatically

boba_community/hc-captcha-metafaucet/api/hc_verifyCAPTCHA.py
```diff
@@ -0,0 +1,57 @@
+import json
+import redis
+import os
+from web3 import Web3
+from eth_abi import abi
+import textwrap
+from dotenv import load_dotenv
+load_dotenv()
+
+REDIS_URL = os.environ.get('REDIS_URL')  # is None if not found
```
wsdt, line 10: Redis instance for all 3 lambda function needs to be the same

boba_community/hc-captcha-metafaucet/api/run-local-server-verifyCAPTCHA.py
```diff
@@ -0,0 +1,6 @@
+import sys
+import json
+from hc_verifyCAPTCHA import hc_verifyCAPTCHA
+
+# EVERYTHING you print here will be directly returned to TypeScript
+print(json.dumps(hc_verifyCAPTCHA(json.loads(sys.argv[1]), None)))
```
wsdt, line 6: these two files are needed for inter-process-communication with JavaScript (only for integration tests). 
Won't be needed in production.

boba_community/hc-captcha-metafaucet/contracts/BobaMainnetFaucet.sol
```diff
@@ -0,0 +1,122 @@
...
+            (bool sent,) = (msg.sender).call{gas: SAFE_GAS_STIPEND, value: _amount}("");
+            require(sent, "Failed to send native");
+
+        emit WithdrawNative(msg.sender, _amount);
+    }
+
+    function getNativeFaucet(
+        bytes32 _uuid,
+        string memory _key,
+        address _to
```
wsdt: Uuid & ImageKey are coupled with to Address. 
Thanks @souradeep-das !

wsdt: `to` as separate parameter for meta tx api.

boba_community/hc-captcha-metafaucet/contracts/TuringHelperMock.sol
wsdt, line 1: only used for foundry tests / unit tests. 
Integration tests use real HybridCompute Helper etc

boba_community/hc-captcha-metafaucet/hardhat.config.ts
```diff
@@ -0,0 +1,50 @@
...
+dotenv.config();
+
+const config: HardhatUserConfig = {
+  mocha: {
+    timeout: 300000,
+  },
+  networks: {
+    boba_local: {
+      url: 'http://localhost:8545',
+      accounts: ['0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80', '0xdbda1821b80551c9d65939329250298aa3472ba22feea921c0cf5d620ea67b97', '0xde9be858da4a475276426320d5e9262ecfc3ba460bfac56360bfa6c4c28b4ee0', '0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80', '0xc526ee95bf44d8fc405a158bb884d9d1238d99f0612e9f33d006bb0789009aaa', '0x8166f546bab6da521a8369cab06c5d2b9e46670292d85c875ee9ec20e84ffb61'],
```
wsdt, line 16: local private keys, are public anyway through docker-compose.yml

Added here for convenience.

boba_community/hc-captcha-metafaucet/contracts/BobaFaucet.sol
```diff
@@ -0,0 +1,150 @@
...
+        uint256 result = _verifyKey(_uuid, hashedKey, _to);
+
+        require(result == 1, 'Captcha wrong');
+
+        claimRecords[_to] = block.timestamp;
+
+        (bool sent,) = (_to).call{gas: SAFE_GAS_STIPEND, value: nativeFaucetAmount}("");
+        require(sent, "Failed to send native");
+
+        if (tokenFaucetAmount > 0 && token.balanceOf(address(this)) >= tokenFaucetAmount) {
```
wsdt, line 132: Mainnet faucet won't issue Boba tokens, to also don't break the testnet faucet (when boba tokens are out) we simply don't try to transfer tokens in this case.

boba_community/hc-captcha-metafaucet/contracts/BobaFaucet.sol
```diff
@@ -0,0 +1,151 @@
+//SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.9;
+
+import '@openzeppelin/contracts/access/Ownable.sol';
+import '@openzeppelin/contracts/token/ERC20/IERC20.sol';
+import './TuringHelper.sol';
+
+contract BobaFaucet is Ownable {
```
InoMurko, line 8: did you roll this out yourself or copy from some of the faucets that already exist?

wsdt: It's based on the original HCCaptcha faucet we had (I guess Jan did it before I joined). 

But yes had to made several changes, you can compare the contracts here: 
https://github.com/bobanetwork/boba/pull/1227/files#diff-ce6a41e1d3ee6284d1dce3eb4ee3fca136d43215234e241598a31096e7373886

InoMurko: Is there something that we could use that was already audited? Preferably it doesn't use HC.

wsdt: Could do the Captcha check on the API as well so that onlyOwner can issue funds through gasless tx. 

Should be a smaller change. 

But that would actually increase the difference to what Jan developed back then. 

If you compare Jan's and my version you can see that I lostly just changed method signatures etc. 

Will go through the repo if we have something we could use.

InoMurko: I don't think any was ever audited. So I would prefer if there's something from openzeppelin or other projects that actually went through an audit. Or gnosis chain faucet sol for example

InoMurko: It would also be better if there was no HC.

wsdt: Gnosis chain faucet is not open source for my knowledge. Will re-check if there are other faucets that have been actually audited or already contracts we can use. 


Would you feel better if the faucet is being done without smart contract at all?

wsdt: After looking even through the Github global search, etc.:

I haven't found any open source mainnet faucets that were audited or are live. 
We could though use an existing testnet faucet (if the license is fine), but not seen any of those being audited as well. 

My suggestions: 
- We keep the faucet as is, as it's heavily based on our previous faucet, etc. and limit our exposure as explained below.
- Could further trim down the mainnet faucet and possibly remove HybridCompute, to make an audit as cheap as possible. 
- Could fork a testnet faucet that is widely used (best probably in terms of security would be PoWFaucet, but that one possibly requires some extra frontend work and has higher infra costs)

What we could also do for the mainnet faucet: 
- Only a very small ETH/Boba amount is handled by the smart contract, and we evaluate usage and if an audit makes actually sense so that we can deposit more funds at once. Worst thing in that case that could happen is loosing e.g. 0.1 ETH or whatever we plan to deposit - which would still be cheaper than doing a fully-fledged audit for a faucet (especially if not many people use it).

wsdt:
@InoMurko Suggesting we merge this, as it actually cleans up our deprecated faucet examples and adds a simple example for how to use HybridCompute. 

Alternatively, I suggest we close this PR and I'll create a new one, cleaning up the old faucets.

boba_community/hc-captcha-metafaucet/hardhat.config.ts
```diff
@@ -0,0 +1,50 @@
...
+      accounts: process.env.PRIVATE_KEY !== undefined ? [process.env.PRIVATE_KEY] : [],
+    },
+    boba_bnb_testnet: {
+      url: 'https://boba-bnb-testnet.gateway.tenderly.co/1clfZoq7qEGyF4SQvF8gvI',
+      accounts: process.env.PRIVATE_KEY !== undefined ? [process.env.PRIVATE_KEY] : [],
+    },
+    boba_mainnet: {
+      url: 'https://boba-ethereum.gateway.tenderly.co/1clfZoq7qEGyF4SQvF8gvI',
+    },
+  },
```
InoMurko: are these ENV configurable?

wsdt: added!

InoMurko (approved):

souradeep-das (approved):
🚀

PR #1229, opened 2023-09-26
rename packages for consistency in NPM

InoMurko:
Some packages are named boba (org) other bobanetwork. Because Gateway is decoupling from monorepo we needed to publish packages to NPM. We have published packages previously (org bobanetwork) so now we're renaming the rest for consistency - since contracts, register and others need to be published as well.


sk-enya:
As we can not use `@eth-optimism/sdk` directly from npm we also need to push our copy of this package to npm under name `@bobanetwork/sdk` so we can have the access contract address
https://github.com/bobanetwork/boba/blob/develop/packages/sdk/src/utils/contracts.ts#L53

PR #1300, opened 2023-11-13
feat: Add Light Bridge release pipeline

wsdt:
## Overview

Since we switch over to DockerHub, I created this pipeline to automate it just as we did for the Faucet.

Workflow is only triggered on release or when we want to trigger it specifically. 
And will be aborted if there are no changes in the teleportation folder. 

Also extending this for an actual aws deployment in a separate pr.

## Changes

- Add Github workflow for light bridge Docker build
- Remove orphaned code
- Add optional env to ops/docker-compose*-side.yml

## Testing

Fully tested in a personal repository.


packages/boba/teleportation/src/exec/run.ts
```diff
@@ -66,10 +66,6 @@ const main = async () => {
     'teleportation-airdrop-gas-amount-wei',
     env.TELEPORTATION_AIRDROP_GAS_AMOUNT_WEI || '100000000000000' // 0.0001 eth
   )
-  const TELEPORTATION_AIRDROP_MIN_USD_VALUE = config.str(
```
wsdt, line 69: was removed in an earlier pr a while ago, cleanup

souradeep-das (approved):

.github/workflows/light-bridge-release.yml
```diff
@@ -0,0 +1,66 @@
+name: Build & Release Light Bridge
```
InoMurko: is there a reason why light bridge has it's own pipeline?
we're already building and pushing containers as part of other pipeline

wsdt: Not sure if we actually want to do a release of light bridge with every regular release. but I can move it into our regular pipeline.

wsdt: done!

InoMurko (approved):

PR #1373, opened 2024-03-26
Add ETH dump

boyuan-chen:

## Overview

Add ethdump 

## Changes

- Add ethdump function when we change the ETH balance and mint ETH on L2

## Testing

The unit test is added

### DIFF
--- a/l2geth/ethdumper/dumper.go
+++ b/l2geth/ethdumper/dumper.go
@@ -42,7 +42,7 @@ func NewEthDumper() EthDumper {
 		}
 	}
 	ethAddres := EthAddress{}
-	err = json.Unmarshal([]byte(byteFile), &ethAddres)
+	err = json.Unmarshal(byteFile, &ethAddres)
 	if err != nil {
 		panic(err)
 	}
--- a/l2geth/ethdumper/dumper.go
+++ b/l2geth/ethdumper/dumper.go
@@ -1,0 +1,99 @@
+package ethdumper
+
+import (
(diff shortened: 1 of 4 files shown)


l2geth/ethdumper/dumper.go
```diff
@@ -0,0 +1,99 @@
...
+
+	s.ethAddress.Addresses = append(s.ethAddress.Addresses, address)
+
+	log.Info("Found eth address", "address", address, "total", len(s.ethAddress.Addresses))
+
+	content, err := json.Marshal(s.ethAddress)
+	if err != nil {
+		panic(err)
+	}
+	err = os.WriteFile(os.Getenv("L2GETH_ETH_DUMP_PATH"), content, 0644)
```
mmontour1306, line 81: It looks like this is overwriting the entire file each time a new address is seen. That seems inefficient, but I suppose it doesn't matter if the total dataset is small enough. I'll approve the overall PR, but if you do see performance issues then consider appending to a file one address at a time.

mmontour1306 (approved):

l2geth/ethdumper/dumper.go
boyuan-chen: Yes, if we change how the file is written, we'll need to make some changes in boba-chain-ops.

jyellick (approved):
Agreed with respect to some of the concerns around rewriting this JSON file each time, but as a one off, if it works, great.

PR #1379, opened 2024-04-09
docs: add banner for fee token switching

souradeep-das:
adds banner indicating fee token support will be through AA


wsdt (approved):

jyellick:
Do you think it makes sense to add this banner into the legacy repo? The fact that it's in the legacy would seem to imply that already.  For the HC, I added an intermediary 'placeholder' page where I stuck the banner, and then a link back to the old doc. https://github.com/bobanetwork/boba/pull/161 What do you think?

souradeep-das:
makes sense, will open one on the main repo!

PR #1380, opened 2024-04-11
Update fraud-detector doc for Anchorage

mmontour1306:
## Overview

Updates legacy fraud-detector README for Anchorage.

### DIFF
--- a/boba_community/fraud-detector/README.md
+++ b/boba_community/fraud-detector/README.md
@@ -2,7 +2,9 @@
 description: Learn how to help detect operator fraud
 ---
 
-# Checking Boba Mainnet for Fraud
+# Checking Boba Mainnet for Fraud (pre-Anchorage)
+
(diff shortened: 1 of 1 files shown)


jyellick (approved):

boba_community/fraud-detector/README.md
```diff
@@ -2,7 +2,9 @@
 description: Learn how to help detect operator fraud
 ---
 
-# Checking Boba Mainnet for Fraud
+# Checking Boba Mainnet for Fraud (pre-Anchorage)
```
jyellick, line 5: Do we think it makes sense to mention Anchorage in the legacy repo?

What we've done for other pages which are outdated but linked to from the doc is to either delete the link, or, add an intermediary landing page which mentions that the old doc is pre-anchorage.  In this case I would kind of think simply deleting the link might make the most sense?

PR #1384, opened 2024-04-15
chore: Verify contracts docs

wsdt:
# Changes
Verify contracts docs updated.


wsdt:
looks like docs moved to new boba repo already

